PR #9, opened 2021-03-10
Bootstrapping the Django app

### LINKED ISSUE #8: Bootstrapping the app
Adding a bootstrap.sh file to automatically bring up the Django environment with one command.
After that the command 'vagrant up' will:
- Install the pipenv tool
- tell pipenv to install the dependencies 
- start the app

ellaml:
fix #8 
Adding a bootstrap.sh file to automatically setup the Django development
environment:
- Setup [Pipenv][1] to install and manage Django and other Python
dependencies
- Bootstrapped an empty [Django][2] application
- Setup Vagrant to automatically bring our application up as well as
make it accessible from a web browser

After running `vagrant up` the application can be accessed at:
> http://127.0.0.1:8000/

**Note:** Most changes had been auto-generated, only `Vagrantfile` and
`bootstrap.sh` had been edited manually.

[1]: https://github.com/pypa/pipenv
[2]: https://www.djangoproject.com/

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash 
+
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+pipenv sync --dev
+
+# The app logs will be collected in nohup.out
+nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,12 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+django = "*"
+
+[requires]
+python_version = "3.8"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,50 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "99c4b9ec1b8891ff787677276760beb6d6d4919c55660da1c713682156a6086c"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.8"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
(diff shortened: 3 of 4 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,15 @@
...
+# password if we don’t have sufficient privileges to run it
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+# Install dependencies with Pipenv
+pipenv sync --dev
+
+# run our app. Nohup and “&” are used to let the setup script finish
+# while our app stays up. The app logs will be collected in nohup.out
+nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
```
Yarboa: See mark for missing new line at the end

bootstrap.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -ex
```
lmilbaum: I would suggest you use the following syntax:
`#!/usr/bin/env bash`

bootstrap.sh
```diff
@@ -0,0 +1,15 @@
...
+# The -e option would make our script exit with an error if any command
+# fails while the -x option makes verbosely it output what it does
+
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don’t have sufficient privileges to run it
```
lmilbaum: Please remove all those comments. Comments are there only to explain why the code was implemented in a certain way.

lmilbaum (changes requested):

bootstrap.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash 
+
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+# Install dependencies with Pipenv
```
lmilbaum: Why is this comment needed? The code should be self explanatory.

lmilbaum (changes requested):
Is there a need for 3 commits? Squash them if not.

lmilbaum (approved):

PR #11, opened 2021-03-10
Fixing the corrupted manage.py file

### LINKED ISSUE #10: Manage.py script is "corrupted"
When Manage.py was pulled from our main repository, All of its "LF"(Linux) signs were replaced with "CRLF" (Windows) signs.
This caused some issues when trying to use it on the Linux VM.
This git behavior is caused due to a default configuration of git for our windows machines.

### 
**To Fix this issue, we need to:**
1. Change the git global config file which is located in: C:\Program Files\Git\etc\gitconfig
   `autocrlf = true` (the default value) should be changed to `autocrlf = false`

2. Run the following commands to change the value to false on your working dir:
   `git config --global core.autocrlf false`   ( per-user solution )
   `git config --local core.autocrlf false `      ( per-project solution )

**To make sure this fix worked**:
Use the command "git config core.autocrlf" in the working dir to check the currently used value of autocrlf*

You can read more about this issue in the following article:
https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf

LeonRabinovich:
The current manage.py file in our repo has "CRLF" signs instead of "LF" signs.
More details on this issue: fix #10

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+###!/usr/bin/env python
 """Django's command-line utility for administrative tasks."""
 import os
 import sys
--- a/manage.py
+++ b/manage.py
@@ -1,4 +1,4 @@
-###!/usr/bin/env python
+#!/usr/bin/env python
 """Django's command-line utility for administrative tasks."""
 import os
 import sys


manage.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+###!/usr/bin/env python
```
lmilbaum: There are too many '#' here. One is enough.

lmilbaum (changes requested):

manage.py
LeonRabinovich: Fixed

lmilbaum (changes requested):
There are not files changed in this PR and two commits. Something is off.

PR #22, opened 2021-03-23
Adding the Course Model

### LINKED ISSUE #18: Create Course Model and Migrations

VeraTs:
Added a Curse Model according to the DB ERD, and manually added to the migrations 0001_initial file.

Currently contains the fields:
- id
- name
- mandatory
- credit_points
- syllabi (is supposed to be a link in the future, for now placeholder String type)
- avg_rating
- avg_load
- num_of_raters (number of raters)
- num_of_reviewers (number of reviewers)

close #18

### DIFF
--- a/homepage/migrations/0001_initial.py
+++ b/homepage/migrations/0001_initial.py
@@ -16,7 +16,15 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='Course',
             fields=[
-                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('course_id', models.IntegerField(primary_key=True)),
+                ('name',  models.CharField(max_length=100)),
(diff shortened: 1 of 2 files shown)


ellaml (approved):

oriAdler (commented):
Looks good! 
Maybe the mandatory field can be BooleanField?

VeraTs:
If I made it Boolean it would work for django, but be relegated to NUMERIC affinity in SQLite, which is a bit ambiguous to my tastes, so I went with smallInt.
(because BOOLEAN would be a NUMERIC affinity but converted to INTEGER if i use Boolean. In that case I though making it an integer from the get go is better)

oriAdler (approved):
LGTM

lmilbaum (changes requested):
Issue is not linked

VeraTs:
linked to issue!

lmilbaum (approved):

PR #29, opened 2021-03-31
Adding a migration test file for the user data model

LeonRabinovich:
The test file contains 3 users which will be added to our DB after running `pipenv run python manage.py migrate`

The Users:
1.  **Name =** 'testUser1', **Email =** 'user1@gmail.com', **Password =**'password123',
2.  **Name =**'testUser2', **Email =**'David@gmail.com', **Password =**'Password456',
3.  **Name =**'testUser3', **Email =**'User3@gmail.com', **Password =**'User3Password'

Resolves: #28

PR #35, opened 2021-04-08
Course and Prerequisites test data

### LINKED ISSUE #31: Create migrations file for Course test data
Include at least 3 courses

### LINKED ISSUE #34: Create all Test data required - contains file names for clarity
0002: User table
0003: Course + Prerequisites table
0004: Professor + prof-to-course tables - dependent on 0003
0005: Review table - dependent on 0002, 0003

VeraTs:
Uses the name 0003_course_test_data
contains:
- course test data
- prerequisites test data

**the courses are:** (id and name only excerpt)
10111 | Resonance in Runes and Signs
10221 | Grammatica in Arithmancy
10231 | UnFogging the Future
12357 | Numerology

** the prerequisites are:**
1 | 12357 | 10231 | 1
2 | 10111 | 10221 | 0

ref #34 fixes #31

### DIFF
--- a/homepage/migrations/0003_course_test_data.py
+++ b/homepage/migrations/0003_course_test_data.py
@@ -1,0 +1,36 @@
+from django.db import migrations, transaction
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('homepage', '0001_initial'),
+    ]
+
+    def generate_data(apps, schema_editor):
+        from homepage.models import Course
+        from homepage.models import Prerequisites
+
+        test_data_course = [
+            (10231, 'UnFogging the Future', 1, 4, 'Currently Unavailable', 3.25, 1.5, 12, 12),
+            (10221, 'Grammatica in Arithmancy', 0, 3, 'Currently Unavailable', 4.25, 4.5, 10, 9),
+            (10111, 'Resonance in Runes and Signs', 1, 3, 'Currently Unavailable', 2.25, 2.0, 13, 13),
+            (12357, 'Numerology', 0, 5, 'Currently Unavailable', 3.5, 2.5, 7, 3),
+        ]
+
+        courses = [Course(*tdc) for tdc in test_data_course]
+        test_data_prerequisites = [
+            (courses[3], courses[0], 1),
+            (courses[2], courses[1], 0),
+            ]
+
+        with transaction.atomic():
+            for course in courses:
+                course.save()
+
+            for preq in test_data_prerequisites:
+                Prerequisites(course_id=preq[0], req_course_id=preq[1], req_code=preq[2]).save()
+
+    operations = [
+        migrations.RunPython(generate_data),
+    ]


homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+        test_data_course = [
+            (10231, 'UnFogging the Future', 1, 4, 'Currently Unavailable', 3.25, 1.5, 12, 12),
+            (10221, 'Grammatica in Arithmancy', 0, 3, 'Currently Unavailable', 4.25, 4.5, 10, 9),
+            (10111, 'Resonance in Runes and Signs', 1, 3, 'Currently Unavailable', 2.25, 2.0, 13, 13),
+            (12357, 'Numerology', 0, 5, 'Currently Unavailable', 3.5, 2.5, 7, 3),
+        ]
+
+        courses = [Course(course_id=t[0], name=t[1], mandatory=t[2],
+                          credit_points=t[3], syllabi=t[4], avg_rating=t[5],
+                          avg_load=t[6], num_of_raters=t[7], num_of_reviewers=t[8]) for t in test_data_course]
```
kobihk: please try to prevent using one char variable name like "t" and use meaningful name even acronym its better like tdc

kobihk: btw in this case you can do it more pythonic way like:
        courses = [Course(*tdc) for tdc in test_data_course]

for more details you can see for example:
https://www.geeksforgeeks.org/args-kwargs-python/

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+
+        prerequisite_1 = Prerequisites(course_id=courses[3], req_course_id=courses[0], req_code=1)
+        prerequisite_2 = Prerequisites(course_id=courses[2], req_course_id=courses[1], req_code=0)
```
kobihk: maybe you can do the same as you did in lines # 14-23 like:
prerequisites = [
    (courses[3], courses[0], 1),
     (courses[2], courses[1], 0),
]
...

in this way you also can easily add/remove prerequisites

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+
+        with transaction.atomic():
+            for course in courses:
+                course.save()
+
+            prerequisite_1.save()
+            prerequisite_2.save()
```
kobihk: same ^^^

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,38 @@
+from django.db import migrations, transaction
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('homepage', '0001_initial'),
```
kobihk, line 7: maybe it will be good to update it after 
https://github.com/beyond-io/ClassRater/pull/29 will merge

kobihk (changes requested):

homepage/migrations/0003_course_test_data.py
VeraTs: Thank you for the feedback! 
Changed this part as you've said, and I'll try to use more meaningful names in the future

homepage/migrations/0003_course_test_data.py
VeraTs: For the user downloading this repository, yes, but this file is independent of the user file completely
I could change the numbering to 0002 to discourage loss of data \ manage.py ignoring the middle migration, if that would work better?
But then #29 would have to change the number as well

VeraTs:
changed and recommitted - still under the 0003 name.

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,37 @@
...
+        from homepage.models import Prerequisites
+
+        test_data_course = [
+            (10231, 'UnFogging the Future', 1, 4, 'Currently Unavailable', 3.25, 1.5, 12, 12),
+            (10221, 'Grammatica in Arithmancy', 0, 3, 'Currently Unavailable', 4.25, 4.5, 10, 9),
+            (10111, 'Resonance in Runes and Signs', 1, 3, 'Currently Unavailable', 2.25, 2.0, 13, 13),
+            (12357, 'Numerology', 0, 5, 'Currently Unavailable', 3.5, 2.5, 7, 3),
+        ]
+
+        courses = [Course(*tdc) for tdc in test_data_course]
```
kobihk, line 21: I think you can remove this line and change the loop in line # 28

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,37 @@
...
+        test_data_prerequisites = [
+            (courses[3], courses[0], 1),
+            (courses[2], courses[1], 0),
+            ]
+
+        with transaction.atomic():
+            for course in courses:
+                course.save()
```
kobihk, line 29:
```suggestion
            for tdc in test_data_course:
                Course(*tdc).save()
```

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+            for preq in test_data_prerequisites:
+                Prerequisites(course_id=preq[0], req_course_id=preq[1], req_code=preq[2]).save()
+
+    operations = [
+        migrations.RunPython(generate_data),
+
```
kobihk: please remove this empty line

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,37 @@
...
+            (courses[3], courses[0], 1),
+            (courses[2], courses[1], 0),
+            ]
+
+        with transaction.atomic():
+            for course in courses:
+                course.save()
+
+            for preq in test_data_prerequisites:
+                Prerequisites(course_id=preq[0], req_course_id=preq[1], req_code=preq[2]).save()
```
kobihk, line 32: its not too much parameters so just if you want you can do it like
```suggestion
                Prerequisites(*preq).save()
```

homepage/migrations/0003_course_test_data.py
kobihk: never mind you can leave it like that
if needed we can change it in other PR

kobihk (changes requested):

homepage/migrations/0003_course_test_data.py
VeraTs: I think I should leave it like that since I need Course objects for the Prerequisites definition - both occurances of Course there are as Foreign Keys, and I can't pass it otherwise

homepage/migrations/0003_course_test_data.py
VeraTs: I tried it when I was fixing this file, but it didn't work then and doesn't work now either, and I'm not sure why, but for some reason this interrupts the Foreign Key part of the definition - I get the error:
`TypeError: Field 'id' expected a number but got <Course: Course object (12357)>.`
If I change the test_data_prerequisites to contain courses[i].course_id it gives another error that indicates that this change prevents correct data placement: it says I have accidentally given a primary key to the instance of Prerequisites.
As such, I'd rather leave it as it is, because it works, and has only 3 parameters.

homepage/migrations/0003_course_test_data.py
VeraTs: removed and recomitted

homepage/migrations/0003_course_test_data.py
kobihk: I'm ok with this code too :)
But...
it looks like the preq[0] is:
            (courses[3], courses[0], 1)
so preq[0] is courses[3] which is Course object that initialized with the following parameters in line # 21:
            (12357, 'Numerology', 0, 5, 'Currently Unavailable', 3.5, 2.5, 7, 3),
so if you want only the course_id you should use it only something like:
test_data_prerequisites = [
            (courses[3].course_id, courses[0].course_id, 1),
            (courses[2].course_id, courses[1].course_id, 0),
            ]

VeraTs: Thank you for replying on the weekend, it's really apriciated! 

I tried exactly that but:
- with the way the prerequisites are saved in the current code I get the error that it expects an object for the foreign key and I'm sending it an integer
- with the Prerequisites(*args) way it somehow overrides the automated PK and puts the last thing in the tuple (0 for example) as the req_course_id which is wrong

I'm really not sure why at all, couldn't find a good enough query for google.

VeraTs: I figured it out, by using this code:
`        test_data_prerequisites = [`
`            (1, courses[3].course_id, courses[0].course_id, 1),`
`            (2, courses[2].course_id, courses[1].course_id, 0),
            ]`
then this works:
`            for preq in test_data_prerequisites:`
                `Prerequisites(*preq).save()`

It means I overrode the auto-PK this is supposed to produce, so I don't know if this is best practice for test data?
If it's okay, I'll commit the new file.

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,36 @@
...
+            (10231, 'UnFogging the Future', 1, 4, 'Currently Unavailable', 3.25, 1.5, 12, 12),
+            (10221, 'Grammatica in Arithmancy', 0, 3, 'Currently Unavailable', 4.25, 4.5, 10, 9),
+            (10111, 'Resonance in Runes and Signs', 1, 3, 'Currently Unavailable', 2.25, 2.0, 13, 13),
+            (12357, 'Numerology', 0, 5, 'Currently Unavailable', 3.5, 2.5, 7, 3),
+        ]
+
+        courses = [Course(*tdc) for tdc in test_data_course]
+        test_data_prerequisites = [
+            (1, courses[3].course_id, courses[0].course_id, 1),
+            (2, courses[2].course_id, courses[1].course_id, 0),
```
kobihk: sorry I missed that you have id as first param which automatically created. 
so better to back to the first code you did
something like:
        test_data_prerequisites = [
            (courses[3].course_id, courses[0].course_id, 1),
            (courses[2].course_id, courses[1].course_id, 0),
         ]

homepage/migrations/0003_course_test_data.py
```diff
@@ -0,0 +1,36 @@
...
+            ]
+
+        with transaction.atomic():
+            for course in courses:
+                course.save()
+
+            for preq in test_data_prerequisites:
+                Prerequisites(*preq).save()
```
kobihk: maybe here you need to back to the code you did in the beginning like:
                Prerequisites(req_code=preq[0], course_id=preq[1], req_course_id=preq[2]).save()

kobihk (commented):
please rebase

homepage/migrations/0003_course_test_data.py
VeraTs: Ok, recommited and rebased.

VeraTs: And thanks again for reviewing in the weekend and at these hours

kobihk (approved):

PR #46, opened 2021-04-11
Changing the user model in models.py

### LINKED ISSUE #40: Change the user model in models.py
- Deleting our old user class 
- Adding a new user class called "AppUser" which uses the built in Django user class (django.contrib.auth.user == django.conf.settings.AUTH_USER_MODEL) 
- Changing the old user field of the review class ("User") to our new class ("AppUser")

LeonRabinovich:
Our new user class is called "AppUser" and it contains the django built in user class 
(django.contrib.auth.User == django.conf.settings.AUTH_USER_MODEL)

Resolves : #40

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -1,5 +1,6 @@
 from django.db import models
 from django.utils import timezone
+from django.conf import settings
 
 
 class Course(models.Model):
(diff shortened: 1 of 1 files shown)


homepage/models.py
```diff
@@ -40,15 +41,16 @@ class Req_Code(models.IntegerChoices):
...
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+
+    def __str__(self):
+        return self.user.username
 
 
 class Review(models.Model):
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
```
VeraTs, line 53: Didn't we want to prevent deletion if a user added a review?

LeonRabinovich: We agreed that users will not have the ability to delete their account (they can be changed to non-active and will not be able to login anymore if this is something we want and can be easily done)

If we forcefully delete a user, i think it makes sense to also delete all of the user's reviews (over-wise we probably wouldn't forcefully delete his account )

VeraTs: Ok, seems I forgot the agreed implementation. 
In that case - LGTM.

VeraTs (approved):

ellaml (approved):

oriAdler (approved):
The motivation of AppUser class was, on the one hand, to enjoy all the benefits of the Django User class.
On the other hand, to have the option of adding our unique methods in the future.

kobihk (approved):

PR #48, opened 2021-04-11
Adding the toString methods to the classes in models.py

### LINKED ISSUE #42: Add __str__ function to all models
Update **model.py** - add ```__str__``` function to all models

**Note:** After this update, it will be possible to see models names on the Django admin UI

LeonRabinovich:
This PR is dependent on PR #46 
Resolves #42

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -1,5 +1,6 @@
 from django.db import models
 from django.utils import timezone
+from django.conf import settings
 
 
 class Course(models.Model):
@@ -12,6 +13,9 @@ class Course(models.Model):
     avg_rating = models.DecimalField(max_digits=6, decimal_places=5)  # average course rating
     num_of_raters = models.IntegerField()  # number of raters for course ratnig and course load
     num_of_reviewers = models.IntegerField()  # number of reviewers
+
+    def __str__(self):
+        return self.name
 
 
 class Prerequisites(models.Model):
@@ -39,26 +43,45 @@ class Prerequisites(models.Model):
     # requirement code : depiction of the relationship between A and B as stated above
     req_code = models.SmallIntegerField(choices=Req_Code.choices, default=Req_Code.NONE)
 
+    def __str__(self):
+        return f'req_cour = {self.req_course_id}, desired_cour = {self.course_id}, req_code = {self.req_code}'
 
-class User(models.Model):
-    name = models.CharField(max_length=100)
-    email = models.CharField(max_length=50)
-    password = models.CharField(max_length=30)
+
+class AppUser(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+
+    def __str__(self):
+        return self.user.username
 
 
 class Review(models.Model):
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
     rate = models.SmallIntegerField()
     date = models.DateTimeField(default=timezone.now)
     content = models.TextField()
     course_load = models.SmallIntegerField()
 
+    def __str__(self):
+        MAX_WORDS_PREVIEW = 5
+        MAX_LENGTH_PREVIEW = 40
+
+        shortened_review = ' '.join(self.content.split()[:MAX_WORDS_PREVIEW])
+        shortened_review = shortened_review[:MAX_LENGTH_PREVIEW]
+
+        return f'Preview: {shortened_review}...'
+
 
 class Professor(models.Model):
     name = models.CharField(max_length=100)
+
+    def __str__(self):
+        return self.name
 
 
 class Professor_to_Course(models.Model):
     professor_id = models.ForeignKey(Professor, on_delete=models.CASCADE)
     course_id = models.ForeignKey(Course, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f'prof = {self.professor_id}, cour = {self.course_id}'


homepage/models.py
```diff
@@ -40,25 +44,35 @@ class Req_Code(models.IntegerChoices):
...
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
     rate = models.SmallIntegerField()
     date = models.DateTimeField(default=timezone.now)
     content = models.TextField()
     course_load = models.SmallIntegerField()
 
+    def __str__(self):
+        return 'Preview: {}'.format(self.content[:20])
```
VeraTs: Is it not critical if it cuts off a word in the middle?
If it is, can [this](https://stackoverflow.com/questions/31978086/set-a-limit-to-the-number-of-words-or-characters-in-a-string) be implemented?

LeonRabinovich: From what i understand - you mean cutting the string into segments of words so the word will not be cut in the middle?

Sounds like a good idea, i changed the method and squashed the previous commit.

VeraTs (approved):
LGTM :)

homepage/models.py
VeraTs: Exactly :)

homepage/models.py
```diff
@@ -40,25 +44,47 @@ class Req_Code(models.IntegerChoices):
...
+        MAX_WORDS_PREVIEW = 5
+        wordsCounter = 1
+
+        for word in review_asList:
+            shortened_review = shortened_review + word + ' '
+            wordsCounter += 1
+
+            if(wordsCounter > MAX_WORDS_PREVIEW):
+                break
+
```
ellaml, line 71: Python provides shorter ways to get x words from a string, even the one in link from @VeraTs  's review - 
shortened_review = ' '.join(content.split()[:5]).
Maybe it's better to use it to avoid all these lines?

oriAdler: I agree with @ellaml, this function should be implemented in a more pythonic way

homepage/models.py
```diff
@@ -40,25 +44,47 @@ class Req_Code(models.IntegerChoices):
...
+    def __str__(self):
+        return self.name
+
 
 class Professor_to_Course(models.Model):
     professor_id = models.ForeignKey(Professor, on_delete=models.CASCADE)
     course_id = models.ForeignKey(Course, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return 'p = {}, c = {}'.format(self.professor_id, self.course_id)
```
oriAdler: I think prof and cour will be more appropriate names than p and c

homepage/models.py
```diff
@@ -40,25 +44,47 @@ class Req_Code(models.IntegerChoices):
     req_code = models.SmallIntegerField(choices=Req_Code.choices, default=Req_Code.NONE)
```
oriAdler, line 45: Missing ```__str__``` implementation for Prerequisites class

homepage/models.py
VeraTs: I also agree with @ellaml

homepage/models.py
VeraTs: when you write it, don't miss creating an `__str__` method that references the req_code as a word:
it was intended to be as written in the class Req_code only SIMU (0) is short for simultaneous.

homepage/models.py
VeraTs: agree

homepage/models.py
LeonRabinovich: Agreed, Changed.

homepage/models.py
LeonRabinovich: Changed.

homepage/models.py
LeonRabinovich: Yes,  Deliberately. In my opinion it is not needed for this class.
if we want to add it, it can be done later.
If the purpose is the Django admin page - it is not necessary since we will add the "table view" for all models. 

However, if you think it is necessary - i can add it.
@ellaml @oriAdler @VeraTs ?

homepage/models.py
```diff
@@ -40,25 +44,37 @@ class Req_Code(models.IntegerChoices):
...
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
     rate = models.SmallIntegerField()
     date = models.DateTimeField(default=timezone.now)
     content = models.TextField()
     course_load = models.SmallIntegerField()
 
+    def __str__(self):
+        MAX_WORDS_PREVIEW = 5
+        shortened_review = ' '.join(self.content.split()[:MAX_WORDS_PREVIEW])
```
kobihk, line 69: if my review will be long string without spaces like 'a'*10000 so your "shorten_review" will be not that short... :)
IMHO it will be better to limit it by string len of the string and add tree dots in the end "..."

homepage/models.py
```diff
@@ -40,25 +44,37 @@ class Req_Code(models.IntegerChoices):
...
+        return 'Preview: {}'.format(shortened_review)
```
kobihk: with python3 we have a great format style like:
```suggestion
        return f'Preview: {shortened_review}'

```

homepage/models.py
```diff
@@ -40,25 +44,37 @@ class Req_Code(models.IntegerChoices):
...
+    def __str__(self):
+        return self.name
+
 
 class Professor_to_Course(models.Model):
     professor_id = models.ForeignKey(Professor, on_delete=models.CASCADE)
     course_id = models.ForeignKey(Course, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return 'prof = {}, cour = {}'.format(self.professor_id, self.course_id)
```
kobihk: please use the python3 style it looks better and more readable:

```suggestion
        return f'prof = {self.professor_id}, cour = {self.course_id}'
```

homepage/models.py
oriAdler: @VeraTs I'm not sure I understand, I suggested adding the ```__str__``` method only to the Prerequisites class.
right now when I print a Prerequisites object through the consol it prints ```<Prerequisites: Prerequisites object (1)>```. @LeonRabinovich Anyway, an ```__str__``` method is nice to have not a must. Can be added for uniformity with the rest of the classes.

oriAdler (approved):

homepage/models.py
VeraTs: Prerequisites depicts an important relationship, and I think we'd prefer it to be legible in the shell
And I thought it might be nicer to have an `__str__` method for the req_code as well for the oop and then calling it from the `__str__` method for Prerequisites. 
It doesn't have to happen, as it's the only place where it would be called, and we can manage well enough with just one for Prerequisites.

kobihk (changes requested):

homepage/models.py
LeonRabinovich: @VeraTs 
I Added the __str__ implementation for the Prerequisites class. ( i will commit it in a few minutes )
However, i had some issues to implement this in the Req_Code class without making changes to the class itself - so i decided not to at this point.

If you think its useful,  i suggest that after this PR is merged - open a separate PR just for the Req_Code class.

homepage/models.py
LeonRabinovich: Changed it

homepage/models.py
LeonRabinovich: Changed

homepage/models.py
LeonRabinovich: We discussed earlier and decided we didn't want words to be cut in the middle of the preview. (hence the MAX_WORDS)
I now implemented a MAX_LEN in addition to the MAX_WORDS. 
This should solve both problems. (Also added "...")

VeraTs (approved):

homepage/models.py
ellaml: @LeonRabinovich cutting by length after splitting by words doesn't solve the first problem (words will still be cut in the middle),
though in my opinion cutting words in the middle is completely ok for a preview if "..." is added.

kobihk (approved):

kobihk (changes requested):
@LeonRabinovich please resolve the conflicts

LeonRabinovich:
> 
> 
> @LeonRabinovich please resolve the conflicts

For some reason i didn't see the conflict from my local git repo.
As a workaround, i fixed the conflict manually on github and it seems fine now.

kobihk (approved):

PR #49, opened 2021-04-12
Update Review model

### LINKED ISSUE #44: Update Review model
Add the following fields to the Review model:
  - ```like_num``` (Integer)
  - ```professor_id``` (FK)
  - ```attachment``` (Image)

oriAdler:
Add to ```models.py``` three new fields:
  - likes_num: the number of likes for this review.
  - professor: enable the user to tag the course's professor
  - image: enable the user to upload histogram/meme

Install **Pillow** on the machine to use models.ImageField:
  - ```Pipfile``` and ```Pipfile.lock``` were updated automatically
  - update ```bootstrap.sh``` to install **Pillow**
  on ```vagrant up```

Modify ```setting.py``` to enable image uploading:
  - add urls to **media** folder

Add ```media/images``` folder and a ```histogram.png``` as example

**Notes:**
  - After the merge, the user will be able to add an image to a review
  - Storing the images on the repo is a temporal solution until
  storage on a remote server will be possible

resolves: #44


Pipfile.lock
```diff
@@ -64,17 +103,17 @@
...
-                "sha256:0d94e0e05a19e57a99444b6ddcf9a6eb2e5c68d3ca1e98e90707af8152c90a92",
-                "sha256:35b2d75ee967ea93b55750aa9edbbf72813e06a66ba54438df2cfac9e3c27fc8"
+                "sha256:7893783d01b8a89811dd72d7dfd4d84ff098e5eed95cfa8905b22bbffe52efc3",
+                "sha256:f5bc8ecabc05bb9d291eb5203d6810b49040f6ff446a756326104746cc00c1db"
             ],
-            "version": "==2.2.0"
+            "version": "==2.3.1"
         }
     }
 }
```
oriAdler, line 119: @kobihk As all these changes were automatically updated, I'm not sure if this affects our application in some way?

LeonRabinovich:
Regarding the new warning you receive - it also happened to me after installing a new Django package.

I found this article
https://stackoverflow.com/questions/66971594/auto-create-primary-key-used-when-not-defining-a-primary-key-type-warning-in-dja

_You have unintentionaly updated Django to 3.2 which warns you and as hint text suggest you have to set DEFAULT_AUTO_FIELD as documented This way you avoid unwanted migrations in future releases as value of DEFAULT_AUTO_FIELD will be changed to BigAutoField_

**You should set DEFAULT_AUTO_FIELD explicitly to current DEFAULT_AUTO_FIELD value**

    DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'

ClassRater/settings.py
```diff
@@ -119,3 +119,6 @@
 # https://docs.djangoproject.com/en/3.1/howto/static-files/
 
 STATIC_URL = '/static/'
+MEDIA_URL = '/media/'
+
+MEDIA_ROOT = 'homepage/media'
```
oriAdler, line 124: Is this is the right way to set the **MEDIA_ROOT**, because it's hard-coded I wasn't sure? Is there a way to make it more general?

homepage/models.py
```diff
@@ -46,17 +46,20 @@ class User(models.Model):
...
     date = models.DateTimeField(default=timezone.now)
     content = models.TextField()
     course_load = models.SmallIntegerField()
-
-
-class Professor(models.Model):
-    name = models.CharField(max_length=100)
+    likes_num = models.SmallIntegerField(default=0)
+    professor_id = models.ForeignKey(Professor, on_delete=models.CASCADE, null=True, blank=True)
+    image = models.ImageField(null=True, blank=True, upload_to="images/")
```
oriAdler, line 73: I changed the name of the field to "image" instead of "attachment". Because I believe users will upload a histogram as a screenshot and not as a file, like in the forum on Facebook. Let me know if it makes sense

LeonRabinovich: I think its fine for now to allow only images. (its even better to keep it simple)
 
Big maybe - we can add support for PDFs later (this will probably be a different class field) - it is not essential anyways.

LeonRabinovich:
@oriAdler 
Note that you have a conflict (probably after my PRs were merged)

I had the same with my previous PR and had to manually fix it in github since i couldn't see the conflict in git.

ClassRater/settings.py
VeraTs: It looks ok to me, but how does the url look when checking it out through `127.0.0.1:8000`?

oriAdler:
Thank you @LeonRabinovich, conflicts solved.

ClassRater/settings.py
oriAdler: To be honest, I'm not sure. I didn't change the urls.py file or tried to access it from the frontend.
I only uploaded an image via Django admin UI so far.

VeraTs: I think you should try to change the urls.py and try to access it, according to [this](https://stackoverflow.com/questions/5517950/django-media-url-and-media-root), so we know the setting is right

homepage/models.py
```diff
@@ -54,13 +54,23 @@ def __str__(self):
...
+
 class Review(models.Model):
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
     user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
     rate = models.SmallIntegerField()
     date = models.DateTimeField(default=timezone.now)
     content = models.TextField()
     course_load = models.SmallIntegerField()
+    likes_num = models.SmallIntegerField(default=0)
+    professor_id = models.ForeignKey(Professor, on_delete=models.CASCADE, null=True, blank=True)
```
VeraTs: can I recommend changing the name to `professor` instead of `professor_id`, 
or adding this to the description: ` db_column='professor_id'`
since django automatically adds '_id' to the end of every FK column.

Pipfile.lock
kobihk: if nothing broken so you can keep update it

ClassRater/settings.py
VeraTs: nvm, I see your point from [here](https://docs.djangoproject.com/en/3.2/howto/static-files/#serving-files-uploaded-by-a-user-during-development)

homepage/models.py
oriAdler: Name changed to ```professor```

VeraTs (approved):

ellaml (approved):
LGTM

kobihk:
@oriAdler please update your commit description

oriAdler:
@kobihk commit description was updated

kobihk (approved):

Issue #45, opened 2021-04-11
Update the initial migration file (0001)


LeonRabinovich:
When we finish adding everything to models.py, generate a new initial migration file (`manage.py makemigrations`)

oriAdler:
BTW on this PR we'll have to add ```pipenv run python manage.py migrate``` to **bootstrap.sh** file, It is missing.

PR #50, opened 2021-04-13
Edited Course model and updated migration files

### LINKED ISSUE #43: Update course model
Update **model.py** - change field **mandatory** from ```SmallIntegerField``` to ```Boolean```

### LINKED ISSUE #51: Update migration files and Using updated files with same names
- update the initial migration to reflect changed models correctly
- update test_data migration files to reflect changed models correctly
- note way of using updated migration files if og files exist in local repo

VeraTs:
This update is relevant to current repository state (updated user model but not updated review model yet)
Course model:
    - changed `Course.mandatory` field to be BooleanField
    - added `Course.Details()` method to use in shell
Migration Files:
    - updated 0001_initial to reflect changes made to models
    - updated 0002_user_test_data to work with changes
    - updated in 0001_initial to reflect new Review model and updated order of table columns
    - discarded the update for 0005_review_test_data - not relevant after change of review model

Steps to take to integrate files into local repository, if it exists with prior migrations:
1) Inside the VM:
 - `cd /vagrant`
 - `pipenv run python manage.py --fake homepage 0001`
 - `pipenv run python manage.py flush` and answer `yes`
2) Outside the VM:
 - delete the db from ClassRater folder
3) Inside VM:
 - `pipenv run python manage.py migrate`

Alternatively, you can destroy the VM and recreate it entirely with the new files - don't forget to delete the DB if it's left over.

**Note:** if you have the previous migration files, and you've made changes to an existing model, these might not work for you.
In that case, after pulling the new migration files: do steps 1), 2) and then:

3.1) remove all the updated migration files from the folder, including 0001

3.2) in the VM:
- `pipenv run python manage.py makemigrations`
- `pipenv run python manage.py migrate`

3.3) reenter the test_data files that are unrelated to the model you changed to the migrations folder

3.4) in the VM:
- `pipenv run python manage.py migrate`
and handle any compatibility errors manually

fixes #43, fixes #51 , relevant: #45, #34


oriAdler:
Thank you @VeraTs , your detailed description helped me to solve errors with the migration files and saved me a lot of precious time

oriAdler:
About the ```num_of_raters``` field:
On class Review, at the moment, it is not possible to leave a review without text.
That means that right now, the number of reviewers is equal to the number of raters.
So maybe the ```num_of_raters``` field is redundant?
@VeraTs @ellaml @LeonRabinovich WDYT?

VeraTs:
@oriAdler 
Well, we intended to make leaving an actual review optional, so the distinction should stay, IMO.
Maybe add the `blank=True, null=True` parameters to the definition of that field?
If we do that then according to [this](https://code.djangoproject.com/ticket/9590) we'll get empty strings in the DB, which I'm ok with, and we can update the num_of_reviewers accordingly.

ellaml:
I'm with Vera on this. @oriAdler

oriAdler (approved):

PR #62, opened 2021-04-21
Adding Pytest to the project

### LINKED ISSUE #61: Add Pytest to our project

LeonRabinovich:
The following changes were made:

- Pipefile & Pipefile.lock were updated automatically after installing Pytest in the enviroment
(using `pipenv install --dev pytest` and then `pipenv install --dev pytest-django`)
From what i understand - because they were updated the VM will install pytest and pytest-django autmaticly when we use `vagrant up`

- Created pytest.yml to run our tests automatically on PRs
- Created pytest.ini to configure Django and pyTest settings
- Added a trivial function in tests.py. This will be deleted in my next PR, it was needed to pass all the automatic tests.

Resolves #61

### DIFF
--- a/homepage/tests.py
+++ b/homepage/tests.py
@@ -1,0 +1,3 @@
+def test_trivial():
+
+    assert True
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,16 @@
+name: PyTest
+on: pull_request
+jobs:
+  pytest:
+    name: Run application tests with PyTest
+    runs-on: ubuntu-20.04
+    container: fedora:32
+    steps:
+      - name: Install Pipenv and Git
+        run: dnf install -y pipenv git
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup environment
+        run: pipenv sync --dev
+      - name: Run PyTest
+        run: pipenv run pytest -v
--- a/Pipfile
+++ b/Pipfile
@@ -5,6 +5,8 @@ verify_ssl = true
 
 [dev-packages]
 flake8 = "*"
+pytest = "*"
+pytest-django = "*"
 
 [packages]
 django = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "477d6406ff67cf0d440da6be397b86d7d4c64a19807bcd8ad3501cb6972f8be1"
+            "sha256": "ba885f561ebc7291a060449a889c44a74afb44c2508c341e0f6e9dc144a2416b"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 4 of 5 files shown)


LeonRabinovich:
After many attempts and research - i could not find a way to successfully run pyTest in my environment.
from what i see, pyTest is running on the PR, but it displays an error similar to the one i am receiving.

**I'm using the command `pipenv run python -m pytest -v`:**
image


**If i delete the file "pyTest.ini**" (where DJANGO_SETTINGS_MODULE is referenced), and then run `pipenv run python -m pytest -v`,
I get no errors - but obviously pyTest cannot detect the test files and nothing happens:
image



Help will be appreciated.

bootstrap.sh
```diff
@@ -6,6 +6,7 @@ cd /vagrant
 pipenv sync --dev
 pipenv install Pillow
 pipenv run python manage.py migrate
+pipenv install --dev pytest
```
VeraTs: if you used the following:
`pipenv install --dev pytest`
`pipenv install --dev pytest-django`
then pytest and pytest django are already in the dev tools and are installed by `pipenv sync --dev`
so line 9 is redundant here, I think

VeraTs:
@LeonRabinovich  what you describe looks a bit like what I experienced when I tried running this with new models but old migration files and existing DB by accident. 
Try:
 1. carefully delete DB (or not carefully destroy VM and then delete DB from root folder)
 2. delete all migration files that are not up to date
 3. introduce the up-to-date migration files

LeonRabinovich:
> 
> 
> @LeonRabinovich what you describe looks a bit like what I experienced when I tried running this with new models but old migration files and existing DB by accident.
> Try:
> 
>     1. carefully delete DB (or not carefully destroy VM and then delete DB from root folder)
> 
>     2. delete all migration files that are not up to date
> 
>     3. introduce the up-to-date migration files

I'll try this, thanks

VeraTs:
@LeonRabinovich  Also, try to leave in tests.py the most trivial test
I used this:
image
If this doesn't work the issue is probably with your models.py file

bootstrap.sh
LeonRabinovich: Thanks!
Seems it was not needed.

Not sure i understand though - if we need to manually install it, why we don't need to add it to the bootstrap file?

i mean, why do we need to run on our VM:
`pipenv install --dev pytest`
`pipenv install --dev pytest-django`

but after the Pipefiles are updated - we dont need to run it any more?

LeonRabinovich:
> 
> 
> @LeonRabinovich Also, try to leave in tests.py the most trivial test
> I used this:
> image
> If this doesn't work the issue is probably with your models.py file

Thanks, it did work :) 
I had to re-migrate everything, and also it seems I only installed pyTest and not Django-pyTest.

now eveyrthing works, and the tests passed.

bootstrap.sh
VeraTs: because you update the pipfile once - it's enough. It lives there now. Try destroying your VM and recreating it and see that you can use this, as is.
If I understood correctly, pipfile keeps the developer tools for us and automatically downloads them when called to `sync --dev`.
So as you've added it while using the VM and it lives in the pipfile now, there's no point in denoting this in the bootstrap.sh file

VeraTs (approved):

ellaml (approved):
LGTM

oriAdler (approved):
LGTM

homepage/tests.py
```diff
@@ -0,0 +1,3 @@
+def test_trivial():
```
kobihk, line 1: @LeonRabinovich in the next PR when you add a real tests please don't forget to remove this one

kobihk (approved):

PR #65, opened 2021-04-22
Adding Functions and tests for the User and FollowedUserCourses models

### LINKED ISSUE #56: Add functions + tests for the AppUser and FollowedUserCourses models

LeonRabinovich:
### Edit: I will seperate this into 2 diffrent PRs ( 1 for each class) 
Changed the status to - in progress.
If anyone has any comments in the meantime - it's always appreciated.

----------------------------------------------------------------

Added the following AppUser functions:

- **toggle_user_activation**: toggles the "is_active" attribute of the user
- **create_appUser**: create an appUser and save it in the DB
- **get_all_appUsers**: returns a list of all AppUsers
- **find_appUser**: find an AppUser by username

Added the following FollowedUserCourses  functions:
- **get_courses_followed_by_appUser**: returns a list of all the courses the given AppUser is following

Resolves: #56

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.utils import timezone
 from django.conf import settings
+from django.contrib.auth.models import User
 
 
 class Course(models.Model):
@@ -53,6 +54,34 @@ class AppUser(models.Model):
     def __str__(self):
         return self.user.username
 
+    def toggle_user_activation(self):
+        self.user.is_active = not(self.user.is_active)
+
+    @staticmethod
+    # Creates a new appUser (without superuser permmisions) and saves it in the DB
+    def create_appUser(username, email, password):
+        appUser = AppUser()
+        user = User.objects.create_user(username, email, password)
+        appUser.user = user
+
+        appUser.save()
+        return appUser
+
+    @staticmethod
+    def get_all_appUsers():
+        appUsers = AppUser.objects.all()
+        return list(appUsers)
+
+    @staticmethod
+    def find_appUser(username_ToFind):
+        allAppUsers = AppUser.get_all_appUsers()
+
+        for appUser in allAppUsers:
+            if(appUser.user.username == username_ToFind):
+                return appUser
+
+        return None
+
 
 class FollowedUserCourses(models.Model):
     user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
@@ -60,6 +89,17 @@ class FollowedUserCourses(models.Model):
 
     def __str__(self):
         return f'user = {self.user}, course = {self.course}'
+
+    @staticmethod
+    def get_courses_followed_by_appUser(appUser):
+        allFollowedUserCourses = FollowedUserCourses.objects.all()
+        courses_followed_by_appUser = list()
+
+        for user_course_pair in allFollowedUserCourses:
+            if (user_course_pair.user == appUser):
+                courses_followed_by_appUser.append(user_course_pair.course)
+
+        return courses_followed_by_appUser
 
 
 class Professor(models.Model):
--- a/homepage/tests.py
+++ b/homepage/tests.py
@@ -1,3 +1,69 @@
-def test_trivial():
+import pytest
+from homepage.models import AppUser, FollowedUserCourses, Course
 
-    assert True
+
+@pytest.mark.django_db
+class TestAppUser:
+
+    def test_create_appUser(self):
+        appUser = AppUser.create_appUser('Jhon1', 'Jhon1@mta.ac.il', '1234')
+
+        # check that appUser is the correct type of Object
+        assert isinstance(appUser, AppUser)
+
+        # check that appUser was sucesfully saved in the DB
+        all_AppUsers = AppUser.get_all_appUsers()
+        assert appUser in all_AppUsers
+
+    def test_toggle_user_activation(self):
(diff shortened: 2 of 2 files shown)


ellaml:
Is the username field in the appUser provided by Django a unique field?
If so - does user.objects.createUser() validates that the username provided isn't taken?
Otherwise, I think we should validate it in our create_appUser method, and then raise an error if needed,
and create a test for it.

If we don't need it to be unique then the method "find_by_username" needs to be updated cause it will return only the first one.

LeonRabinovich:
> 
> 
> Is the username field in the appUser provided by Django a unique field?
> If so - does user.objects.createUser() validates that the username provided isn't taken?
> Otherwise, I think we should validate it in our create_appUser method, and then raise an error if needed,
> and create a test for it.
> 
> If we don't need it to be unique then the method "find_by_username" needs to be updated cause it will return only the first one.

Yes, the username is a unique field.
if you try to create 2 users with the same username you will get an exception-

"django.db.utils.IntegrityError: UNIQUE constraint failed: auth_user.username"

homepage/models.py
```diff
@@ -53,6 +54,34 @@ class AppUser(models.Model):
...
+        return appUser
+
+    @staticmethod
+    def get_all_appUsers():
+        appUsers = AppUser.objects.all()
+        return list(appUsers)
+
+    @staticmethod
+    def find_appUser(username_ToFind):
+        allAppUsers = AppUser.get_all_appUsers()
```
oriAdler, line 77: Can I suggest a different approach? ```AppUser.objects.filter(username=username_ToFind)```
Pros:
  - You don't have to retrieve all users.
  - It is faster than iterating over all users.

homepage/models.py
```diff
@@ -61,6 +90,17 @@ class FollowedUserCourses(models.Model):
     def __str__(self):
         return f'user = {self.user}, course = {self.course}'
 
+    @staticmethod
+    def get_courses_followed_by_appUser(appUser):
+        allFollowedUserCourses = FollowedUserCourses.objects.all()
```
oriAdler, line 95: Same here, maybe it is possible to use a query to retrieve courses followed by the user, instead of a ```for``` loop.

homepage/tests.py
```diff
@@ -1,3 +1,69 @@
...
+    def test_toggle_user_activation(self):
+        appUser = AppUser.create_appUser('Jhon2', 'Jhon2@mta.ac.il', '1234')
+
+        appUser.toggle_user_activation()
+        assert appUser.user.is_active is False
+
+        appUser.toggle_user_activation()
+        assert appUser.user.is_active is True
+
+    def test_get_all_appUsers(self):
```
oriAdler, line 27: As ```test_get_all_appUsers``` and ```test_find_appUser``` use almost the same data,
I would suggest using a **fixture**.

homepage/tests.py
```diff
@@ -1,3 +1,69 @@
...
+        user_course_pair_1.save()
+        user_course_pair_2 = FollowedUserCourses(user=appUser1, course=course2)
+        user_course_pair_2.save()
+        user_course_pair_3 = FollowedUserCourses(user=appUser2, course=course2)
+        user_course_pair_3.save()
+
+        appUser1_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser1)
+        appUser2_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser2)
+
+        assert (course1 in appUser1_followed_courses) and (course2 in appUser1_followed_courses)
```
oriAdler, line 68: You can use a third logical ```and``` operator instead of two assert statements

oriAdler (commented):
When separating the PR's, you can add another test for the **FollowedUserCourses** class. For example, a user without followed courses.

homepage/models.py
```diff
@@ -53,6 +54,34 @@ class AppUser(models.Model):
...
+    def toggle_user_activation(self):
+        self.user.is_active = not(self.user.is_active)
+
+    @staticmethod
+    # Creates a new appUser (without superuser permmisions) and saves it in the DB
+    def create_appUser(username, email, password):
+        appUser = AppUser()
+        user = User.objects.create_user(username, email, password)
+        appUser.user = user
+
```
VeraTs, line 66: I think a comment here about what fields we use from the django User for our purpose will be helpful;
for example - that we use the is_active field

VeraTs (commented):
Overall I agree with Ori's comments.

homepage/tests.py
LeonRabinovich: Not sure i understand what you mean by that

oriAdler: My bad, the comment was supposed to be in line 69.
Instead ```assert course2 in appUser2_followed_courses```, add logical ```and (course2 in appUser2_followed_courses)```
Just for uniformity, not a must.

homepage/models.py
LeonRabinovich: not sure what you mean Vera
we can use all of the fields from the built-in Django user class

is_active is automatically set to true because this is the Django beahviour

homepage/models.py
LeonRabinovich: I had some issues using objects.get and objects.filter in this situtation (maybe because we need to do something like `user.username = username_ToFind`)
if you can make it work  - let me know (on the new PR i'm opening)

For now I'm keeping it the old way

PR #66, opened 2021-04-22
Add Review class methods and tests

### LINKED ISSUE #60: Add Review methods and tests
- Add ```print_details``` method
  - Add Review class tests

oriAdler:
Modify **models.py** to add ``print_details`` method

Modify **tests.py** to add **Review** class tests:
  - check the data already stored in the database
  - add a valid review
  - create invalid review (validation constraints)
  - call ```__str__``` method
  - call ```print_details``` method

Modify **0006_review_test_data.py**
Because the migration files are executed upon every test
Check if the test image already exist to avoid duplicates

Add **0007_auto_alter_review_field.py** to modify class fields

fix: #60


homepage/models.py
```diff
@@ -80,10 +81,10 @@ def __str__(self):
 class Review(models.Model):
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
     user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
-    rate = models.SmallIntegerField()
+    rate = models.SmallIntegerField(validators=[MinValueValidator(1), MaxValueValidator(5)])
     date = models.DateTimeField(default=timezone.now)
-    content = models.TextField(null=True, blank=True)
-    course_load = models.SmallIntegerField()
+    content = models.TextField(null=True, blank=True, max_length=200, validators=[MaxLengthValidator(200)])
```
ellaml: I my opinion 200 chars for content is not enough.
TextField is usually used for large texts that don't require max-length limitation, we can just limit the typing in the UI widget. Check out this [post](https://stackoverflow.com/questions/7354588/django-charfield-vs-textfield).
Anyway, if we want to limit it in the DB I think it should be at least 1000 chars.

oriAdler: Good catch!
I'll remove the ```MaxLengthValidator```

homepage/models.py
```diff
@@ -96,3 +97,17 @@ def __str__(self):
         shortened_review = shortened_review[:MAX_LENGTH_PREVIEW]
 
         return f'Preview: {shortened_review}...'
+
+    def print_details(self):
+        message = (
+            f"Course: {self.course}\n"
+            f"User: {self.user}\n"
+            f"Rate: {self.rate}\n"
```
VeraTs: 'Rating' would be better I think

homepage/models.py
```diff
@@ -96,3 +97,17 @@ def __str__(self):
         shortened_review = shortened_review[:MAX_LENGTH_PREVIEW]
 
         return f'Preview: {shortened_review}...'
+
+    def print_details(self):
+        message = (
+            f"Course: {self.course}\n"
+            f"User: {self.user}\n"
+            f"Rate: {self.rate}\n"
+            f"{str(self)}\n"
```
VeraTs: unclear what this is - is this the entire review content?

homepage/tests.py
```diff
@@ -1,3 +1,87 @@
...
+    review.save()
+
+    assert Review.objects.filter(pk=review.id).exists()
+
+
+@pytest.mark.parametrize("invalid_review", [
+    (10, 10231, 3, 10, datetime(2015, 10, 9, 23, 55, 59, 5, tzinfo=pytz.UTC), "Great course!", 4, 0, None, ''),
+    # 0: rate > 5  ^rate
+    (10, 10231, 3, 3, datetime(2015, 10, 9, 23, 55, 59, 5, tzinfo=pytz.UTC), "bla"*100, 4, 0, None, ''),
+    # 1: content.length > 200                                                ^content
```
VeraTs: if you change the content length restriction don't forget to remove this test

VeraTs (changes requested):
Stuff marked

homepage/models.py
oriAdler: It is a part of the review content.
I'll add a comment to clarify

ellaml (approved):

VeraTs (approved):

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +15,39 @@ def generate_data(apps, schema_editor):
...
         professors = Professor.objects.all()
+        image = ''
 
-        image_path = settings.MEDIA_ROOT + '/images/test_image.jpg'
-        image = SimpleUploadedFile(
-            name='new_test_image.jpg',
-            content=open(image_path, 'rb').read(),
-            content_type='image/jpeg')
+        # prevent creating a new image upon every pytest run
+        if not path.exists(settings.MEDIA_ROOT + '/images/new_test_image.jpg'):
```
kobihk: please use path.join it can prevent all the mess of concatenate the path 
```suggestion
        if not path.exists(path.join(settings.MEDIA_ROOT, '/images/new_test_image.jpg')):
```

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +15,39 @@ def generate_data(apps, schema_editor):
...
+            image_path = settings.MEDIA_ROOT + '/images/test_image.jpg'
```
kobihk: same ^^

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +15,39 @@ def generate_data(apps, schema_editor):
...
+            image = SimpleUploadedFile(
+                name='new_test_image.jpg',
+                content=open(image_path, 'rb').read(),
+                content_type='image/jpeg')
+        else:
+            image = 'images/new_test_image.jpg'
```
kobihk: it will be better to create constant like:
new_image_file = 'images/new_test_image.jpg'

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +15,39 @@ def generate_data(apps, schema_editor):
...
+            (2, courses[0].course_id, users[0].user.id, 4, date, "I've learned a lot!", 2, 4),
+            (3, courses[1].course_id, users[1].user.id, 4.5, date,
                 "The course was difficult", 5, 0, professors[0].id),
-            (4, courses[1].course_id, users[1].user.id, 3.2, timezone.now(), "I didn't learn anything new", 3.7),
-            (5, courses[2].course_id, users[2].user.id, 4.5, timezone.now(),
+            (4, courses[1].course_id, users[1].user.id, 3.2, date, "I didn't learn anything new", 3.7),
+            (5, courses[2].course_id, users[2].user.id, 4.5, date,
                 "This course helped me to find a job", 2.5),
-            (6, courses[2].course_id, users[2].user.id, 3.5, timezone.now(), "I didn't understand the material ", 4),
+            (6, courses[2].course_id, users[2].user.id, 3.5, date, "I didn't understand the material ", 4),
```
kobihk:
```suggestion
            (6, courses[2].course_id, users[2].user.id, 3.5, date, "I didn't understand the material", 4),
```

homepage/migrations/0007_auto_alter_review_field.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.2 on 2021-04-25 12:50
+
+import django.core.validators
```
kobihk:
```suggestion
from django.core import validators
```
then you can use it simpler in line # 18

homepage/migrations/0007_auto_alter_review_field.py
```diff
@@ -0,0 +1,26 @@
...
+    dependencies = [
+        ('homepage', '0006_review_test_data'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='review',
+            name='course_load',
+            field=models.SmallIntegerField(
+                validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(5)]),
```
kobihk:
```suggestion
                validators=[validators.MinValueValidator(1), validators.MaxValueValidator(5)]),
```

homepage/migrations/0007_auto_alter_review_field.py
```diff
@@ -0,0 +1,26 @@
...
+        ),
+        migrations.AlterField(
+            model_name='review',
+            name='rate',
+            field=models.SmallIntegerField(
+                validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(5)]),
```
kobihk:
```suggestion
                validators=[validators.MinValueValidator(1), validators.MaxValueValidator(5)]),
```

homepage/models.py
```diff
@@ -96,3 +97,17 @@ def __str__(self):
         shortened_review = shortened_review[:MAX_LENGTH_PREVIEW]
 
         return f'Preview: {shortened_review}...'
+
+    def print_details(self):
+        message = (
+            f"Course: {self.course}\n"
+            f"User: {self.user}\n"
+            f"Rating: {self.rate}\n"
+            f"{str(self)}\n"    # shortened review content
```
kobihk, line 128:
```suggestion
            f"Shortened review: {str(self)}\n"    # shortened review content

```

homepage/models.py
```diff
@@ -96,3 +97,17 @@ def __str__(self):
...
+        message = (
+            f"Course: {self.course}\n"
+            f"User: {self.user}\n"
+            f"Rating: {self.rate}\n"
+            f"{str(self)}\n"    # shortened review content
+            f"Course load: {self.course_load}\n"
+            f"Likes number: {self.likes_num}"
+        )
+        if self.professor is not None:
+            message += f"\nProfessor: {self.professor}"
```
kobihk:
```suggestion
            f"Likes number: {self.likes_num}\n"
            f"Professor: {self.professor if self.professor else 'N/A'}"
        )
```

kobihk (changes requested):

homepage/models.py
oriAdler: I'll change ```def __str__(self)``` last line to ```return f'Shortened review: {shortened_review}...'```
So when calling ```str(self)``` the output will be as expected

LeonRabinovich:
LGTM

LeonRabinovich (approved):

homepage/migrations/0006_review_test_data.py
```diff
@@ -1,4 +1,5 @@
 from django.db import migrations, transaction
+NEW_IMAGE_FILE = 'images/new_test_image.jpg'
```
kobihk: please split it like
```suggestion
IMAGES_DIR = 'images'
IMAGE_FILE = 'test_image.jpg'
NEW_IMAGE_FILE = 'new_test_image.jpg'
```
now you can use it in your code below

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +16,39 @@ def generate_data(apps, schema_editor):
...
+        image = ''
 
-        image_path = settings.MEDIA_ROOT + '/images/test_image.jpg'
-        image = SimpleUploadedFile(
-            name='new_test_image.jpg',
-            content=open(image_path, 'rb').read(),
-            content_type='image/jpeg')
+        # prevent creating a new image upon every pytest run
+        if not path.exists(path.join(settings.MEDIA_ROOT, NEW_IMAGE_FILE)):
+            image_path = path.join(settings.MEDIA_ROOT, 'images', 'test_image.jpg')
```
kobihk:
```suggestion
            image_path = path.join(settings.MEDIA_ROOT, IMAGES_DIR, IMAGE_FILE)
```

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +16,39 @@ def generate_data(apps, schema_editor):
...
+            image = SimpleUploadedFile(
+                name='new_test_image.jpg',
```
kobihk:
```suggestion
                name=NEW_IMAGE_FILE,
```

homepage/models.py
kobihk: good idea

homepage/migrations/0006_review_test_data.py
```diff
@@ -15,30 +18,39 @@ def generate_data(apps, schema_editor):
...
+        if not path.exists(path.join(settings.MEDIA_ROOT, NEW_IMAGE_FILE)):
+            image_path = path.join(settings.MEDIA_ROOT, IMAGES_DIR, IMAGE_FILE)
+            image = SimpleUploadedFile(
+                name=NEW_IMAGE_FILE,
+                content=open(image_path, 'rb').read(),
+                content_type='image/jpeg')
+        else:
+            image = NEW_IMAGE_FILE
+
+        date = datetime(2015, 10, 9, 23, 55, 59, 5, tzinfo=pytz.UTC)
```
kobihk: the "problem" in python is that you can overwrite basic python keywords easily by assign them to something else
for example:
In [1]: str(2)                                                                                                                                                                                                                               
Out[1]: '2'

In [2]: str = 'string'                                                                                                                                                                                                                       

In [3]: str(2)                                                                                                                                                                                                                               
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-531f7a6f46fa> in <module>
----> 1 str(2)

TypeError: 'str' object is not callable

here is not the case but because you imported datetime and it has the "date" method I would suggest to change the name to something less generic 

```suggestion
        cur_date = datetime(2015, 10, 9, 23, 55, 59, 5, tzinfo=pytz.UTC)
```

kobihk (changes requested):

homepage/migrations/0006_review_test_data.py
oriAdler: Thank you for this significant insight. 
I'll be aware of that in the future. 
Changing the variable name.

kobihk (approved):

PR #71, opened 2021-04-26
Add functions and tests for AppUser model

### LINKED ISSUE #69: Add functions and tests for the AppUser model

LeonRabinovich:
Added the following AppUser functions:

- **toggle_user_activation**: toggles the "is_active" attribute of the user
- **create_appUser**: create an appUser and save it in the DB
- **get_all_appUsers**: returns a list of all AppUsers
- **find_appUser**: find an AppUser by username

Resolves: #69

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.utils import timezone
 from django.conf import settings
+from django.contrib.auth.models import User
 
 
 class Course(models.Model):
@@ -48,10 +49,38 @@ class Prerequisites(models.Model):
 
 
 class AppUser(models.Model):
+    # This field is the built-in django user model
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
 
     def __str__(self):
         return self.user.username
+
+    def toggle_user_activation(self):
+        self.user.is_active = not(self.user.is_active)
+
+    @staticmethod
+    # Creates a new appUser (without superuser permmisions) and saves it in the DB
+    def create_appUser(username, email, password):
+        appUser = AppUser()
+        user = User.objects.create_user(username, email, password)
+        appUser.user = user
+
+        appUser.save()
+        return appUser
+
+    @staticmethod
+    def get_all_appUsers():
+        return list(AppUser.objects.all())
+
+    @staticmethod
+    def find_appUser(username_ToFind):
+        allAppUsers = AppUser.get_all_appUsers()
+
+        for appUser in allAppUsers:
+            if(appUser.user.username == username_ToFind):
+                return appUser
+
+        return None
 
 
 class FollowedUserCourses(models.Model):
--- a/homepage/tests.py
+++ b/homepage/tests.py
@@ -1,3 +1,42 @@
-def test_trivial():
+import pytest
+from homepage.models import AppUser
 
-    assert True
+
+@pytest.mark.django_db
+class TestAppUser:
+
+    @pytest.fixture
+    def test_appUsers(self):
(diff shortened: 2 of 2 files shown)


homepage/models.py
```diff
@@ -48,11 +49,39 @@ def __str__(self):
...
+    def toggle_user_activation(self):
+        self.user.is_active = not(self.user.is_active)
+
+    @staticmethod
+    # Creates a new appUser (without superuser permmisions) and saves it in the DB
+    def create_appUser(username, email, password):
+        appUser = AppUser()
+        user = User.objects.create_user(username, email, password)
+        appUser.user = user
+
```
VeraTs, line 67: I still think there should be a comment here explaining which fields from the django user we are using and for what

VeraTs (changes requested):
I think the code looks fine, but it bears commenting. 
You should add a small explanation of what each test is doing and of the fixture you've created.

homepage/models.py
LeonRabinovich: i disagree
I think the code should have as few comments as possible, the code is rather simple and i don't see a section which needs further explaining with comments.

Is there a specific section you think could use more comments?
_______________________________________________
Regarding the Django user fields we are using - currently only username, email password and is_active, but this will probably change in the future.
I will add the current fields we are using though

homepage/models.py
```diff
@@ -48,11 +49,39 @@ def __str__(self):
...
+        appUser.save()
+        return appUser
+
+    @staticmethod
+    def get_all_appUsers():
+        return list(AppUser.objects.all())
+
+    @staticmethod
+    def find_appUser(username_ToFind):
+        allAppUsers = AppUser.get_all_appUsers()
```
oriAdler, line 77: Can you elaborate on which issues you had with using ```filter``` or ```get```?
I still don't see a reason to retrieve all objects and iterate over them
If you can add a screenshot of the error that will be great

homepage/tests.py
```diff
@@ -1,3 +1,42 @@
...
+        appUser2 = AppUser.create_appUser('user2', 'user2@mta.ac.il', '1234')
+        appUser3 = AppUser.create_appUser('user3', 'user3@gmail.com', '10101010')
+
+        return (appUser1, appUser2, appUser3)
+
+    def test_create_appUser(self, test_appUsers):
+        assert isinstance(test_appUsers[0], AppUser)
+
+        all_AppUsers = AppUser.get_all_appUsers()
+        assert test_appUsers[0] in all_AppUsers
```
oriAdler, line 20: Again.. I suggest you use ```assert AppUser.objects.filter(pk=test_appUsers.id).exists()```
instead of retrieving all objects

homepage/models.py
VeraTs: in general - just short, one line comments explaining what each test is testing.
If it is testing valid input, if it is testing invalid input, etc.

homepage/models.py
LeonRabinovich: I was trying to do something like:
AppUser.objects.filter(user.username = username_ToFind)

however its not possible in this way (username is an attribute of User and not AppUser)

After some tests i found another way to do it, so it will be changed in my next commit.

homepage/tests.py
LeonRabinovich: Changed

LeonRabinovich: Changed

LeonRabinovich:
> I think the code looks fine, but it bears commenting.
> You should add a small explanation of what each test is doing and of the fixture you've created.

I added comments where I found necessary.
some tests(functions) are self explanatory by their name - hence IMO comments will only clutter the code and make it less readable.

let me know if after the changes (i will commit in a few minutes) everything is clear

PR #73, opened 2021-04-26
Add functions and tests for FollowedUserCourses model

### LINKED ISSUE #70: Add functions and tests for the FollowedUserCourses model

LeonRabinovich:
Added the following FollowedUserCourses functions:

-   get_courses_followed_by_appUser: returns a list of all the courses the given AppUser is following

Resolves #70

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.utils import timezone
 from django.conf import settings
+from django.contrib.auth.models import User
 
 
 class Course(models.Model):
@@ -53,6 +54,16 @@ class AppUser(models.Model):
     def __str__(self):
         return self.user.username
 
+    @staticmethod
+    # Creates a new appUser (without superuser permmisions) and saves it in the DB
+    def create_appUser(username, email, password):
+        appUser = AppUser()
+        user = User.objects.create_user(username, email, password)
+        appUser.user = user
+
+        appUser.save()
+        return appUser
+
 
 class FollowedUserCourses(models.Model):
     user = models.ForeignKey(AppUser, on_delete=models.CASCADE)
@@ -60,6 +71,17 @@ class FollowedUserCourses(models.Model):
 
     def __str__(self):
         return f'user = {self.user}, course = {self.course}'
+
+    @staticmethod
+    def get_courses_followed_by_appUser(appUser):
+        allFollowedUserCourses = FollowedUserCourses.objects.all()
+        courses_followed_by_appUser = list()
+
+        for user_course_pair in allFollowedUserCourses:
+            if (user_course_pair.user == appUser):
+                courses_followed_by_appUser.append(user_course_pair.course)
+
+        return courses_followed_by_appUser
 
 
 class Professor(models.Model):
--- a/homepage/tests.py
+++ b/homepage/tests.py
@@ -1,3 +1,35 @@
-def test_trivial():
+import pytest
+from homepage.models import AppUser, FollowedUserCourses, Course
 
-    assert True
+
+@pytest.mark.django_db
+class TestFollowedUserCourses:
+
+    def test_get_courses_followed_by_appUser(self):
+        appUser1 = AppUser.create_appUser('user1', 'user1@mta.ac.il', '1234')
+        appUser2 = AppUser.create_appUser('user2', 'user2@gmail.com', '1234')
+        appUser_noCourses = AppUser.create_appUser('user3', 'user3@', '9722')
+
+        course1 = Course(course_id=10231, name='course1', mandatory=True, credit_points=0, avg_load=0,
+                         avg_rating=0, num_of_raters=0, num_of_reviewers=0)
+        course1.save()
+
+        course2 = Course(course_id=10400, name='course2', mandatory=True, credit_points=0,
+                         avg_load=0, avg_rating=0, num_of_raters=0, num_of_reviewers=0)
+        course2.save()
+
+        user_course_pair_1 = FollowedUserCourses(user=appUser1, course=course1)
+        user_course_pair_1.save()
+        user_course_pair_2 = FollowedUserCourses(user=appUser1, course=course2)
+        user_course_pair_2.save()
+        user_course_pair_3 = FollowedUserCourses(user=appUser2, course=course2)
+        user_course_pair_3.save()
+
+        appUser1_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser1)
+        appUser2_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser2)
+        appUser_noCourses_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser_noCourses)
+
+        assert (course1 in appUser1_followed_courses) and (course2 in appUser1_followed_courses)
+        assert course2 in appUser2_followed_courses
+        assert course1 not in appUser2_followed_courses
+        assert appUser_noCourses_followed_courses == []


homepage/models.py
```diff
@@ -61,6 +72,17 @@ class FollowedUserCourses(models.Model):
...
 
+    @staticmethod
+    def get_courses_followed_by_appUser(appUser):
+        allFollowedUserCourses = FollowedUserCourses.objects.all()
+        courses_followed_by_appUser = list()
+
+        for user_course_pair in allFollowedUserCourses:
+            if (user_course_pair.user == appUser):
+                courses_followed_by_appUser.append(user_course_pair.course)
+
```
VeraTs, line 83: As Ori said in the previous PR - please consider using `SomeModel.objects.filter(...)` as it returns a QuerySet and not a list, which seems a more correct way to address a query

homepage/tests.py
```diff
@@ -1,3 +1,35 @@
-def test_trivial():
+import pytest
+from homepage.models import AppUser, FollowedUserCourses, Course
 
-    assert True
+
+@pytest.mark.django_db
+class TestFollowedUserCourses:
+
+    def test_get_courses_followed_by_appUser(self):
```
VeraTs, line 8: A comment explaining what the test does will be great here.
Also, maybe it should be divided into several tests? 
Like:
 - test for user with one followed course
 - test for user with multiple followed courses
 - test for user with no followed courses

VeraTs (commented):
Overall okay, but as I've said in the second PR as well, please add comments that describe what the code is doing

homepage/models.py
LeonRabinovich: The only option in this case is `FollowedUserCourses.objects.filter(user=appUser)`
which unfortunately does not help. 

This command will return a QuerySet which consist of '**FollowedUserCourses**' elements.
Then i still need to iterate other all of the elements in the QuerySet to get only the '**Course**' element

(because the method 'get_courses_followed_by_appUser' needs to return a list of **Course** elements and not **FollowedUserCourses** elements )

homepage/tests.py
LeonRabinovich: Thanks, good idea
I agree that this method is too long, i will split it into smaller methods

homepage/models.py
LeonRabinovich: I did some tests to implement it with objects.filter. 
Let me know if this is clear or if you have ideas for improvements:

image

LeonRabinovich:
> 
> 
> Overall okay, but as I've said in the second PR as well, please add comments that describe what the code is doing

Regarding the comments - i replayed in the second PR.
Also see Liora's message in slack from a while ago -

image

homepage/models.py
oriAdler: @LeonRabinovich you say that you still have to iterate over the elements in the QuerySet, but let me explain the difference.
Let say you have 2000 rows in this table, when you write ```allFollowedUserCourses = FollowedUserCourses.objects.all()```
you're going to retrieve all of them, and iterate over **2000 objects**.
If the user has only 5 followed courses, and you call ```FollowedUserCourses.objects.filter(user=appUser)```, then you have to iterate over **5 objects**.

homepage/tests.py
```diff
@@ -1,3 +1,35 @@
...
+        user_course_pair_2.save()
+        user_course_pair_3 = FollowedUserCourses(user=appUser2, course=course2)
+        user_course_pair_3.save()
+
+        appUser1_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser1)
+        appUser2_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser2)
+        appUser_noCourses_followed_courses = FollowedUserCourses.get_courses_followed_by_appUser(appUser_noCourses)
+
+        assert (course1 in appUser1_followed_courses) and (course2 in appUser1_followed_courses)
+        assert course2 in appUser2_followed_courses
```
oriAdler, line 33: As I mentioned before in the PR you closed
Instead ```assert course2 in appUser2_followed_courses```, add logical ```and (course2 in appUser2_followed_courses)``` to line 32
Just for uniformity, not a must.

homepage/models.py
VeraTs: Update: sorry I misread the intention that you wanted objects and not just string and such. Comment bellow deals with objects.

@LeonRabinovich 
did you consider using this : `list(FollowedUserCourses.filter(user=user).values_list('course'))` to turn the QuerySet into a list?
For example, in my Course Model, I used this static method:
```
    @staticmethod
    def test():
        return list(Course.objects.filter(course_id__contains=1).values_list('name'))
```
and received this output in the shell:
```
>>> res = Course.test()
>>> print(res)
[('Resonance in Runes and Signs',), ('Grammatica in Arithmancy',), ('UnFogging the Future',), ('Numerology',)]
```

**update:**
@ellaml - might be useful for you too? 
After messing with this a little, this method:
```
    @staticmethod
    def test():
        res = [arg[0] for arg in list(Course.objects.filter(course_id__contains=1).values_list('name'))]
        return res
```
return this:
```
>>> res = Course.test()
>>> print(res)
['Resonance in Runes and Signs', 'Grammatica in Arithmancy', 'UnFogging the Future', 'Numerology']
```

homepage/tests.py
VeraTs: regarding the comments - as I've said in the second PR - this is just a request for a short oneliner telling us what you're testing.
For example - the current file has a long test, you've agreed it's good to split it up - and while splitting it into smaller tests will boost readability of the code, a short comment stating what you're testing, above the test method, does not read as clutter to me.
Frankly, if the code had been several tests, I might not have raised this issue if they were testing small enough things, since test names are rather verbose.

homepage/models.py
ellaml: @VeraTs Yes I think I'll use it
The problem began when I thought that maybe I should return a QuerySet instead of the Courses/Pro's.
In that case if I can return a list of objects (which I think is easier to work with for the view) then your code is perfect.

VeraTs: @ellaml  @LeonRabinovich 
The methods I wrote about above give back a list of values, not objects.
If it's a list of objects, then a QuerySet is the best way to go, really. 
For example:
for this function: 
```
    @staticmethod
    def test():
        return Prerequisites.objects.filter(course_id=12357)
```

I get this output:
`<QuerySet [<Prerequisites: req_cour = UnFogging the Future, desired_cour = Numerology, req_code = 1>]>`

and can handle it like so:
```
>>> res[0].course_id
<Course: Numerology>
>>> res[0].course_id.print_details()
```

and print_details() works.

LeonRabinovich: > @LeonRabinovich you say that you still have to iterate over the elements in the QuerySet, but let me explain the difference.
> Let say you have 2000 rows in this table, when you write `allFollowedUserCourses = FollowedUserCourses.objects.all()`
> you're going to retrieve all of them, and iterate over **2000 objects**.
> If the user has only 5 followed courses, and you call `FollowedUserCourses.objects.filter(user=appUser)`, then you have to iterate over **5 objects**.

Yes i agree Ori, however the objects.filter() method probably has to iterate all of the 2000 objects anyways in the background to filter the required objects .. 
it is probably more efficient than a my own iteration though, which is why I did change the method to use filter()

LeonRabinovich: @VeraTs 

Thanks for you help,
I will do some more tests and see if i can write it more efficiently

When I'm done I will commit with all of the changes discussed in the PR

homepage/tests.py
LeonRabinovich: > As I mentioned before in the PR you closed
> Instead `assert course2 in appUser2_followed_courses`, add logical `and (course2 in appUser2_followed_courses)` to line 32
> Just for uniformity, not a must.

Regarding Objects.filter() - see my answer in the thread above (I will be using objects.filter() ) 

Reagarding the second assert - i split it into 2 lines on purpose.
the first line checks courses for appUser1, and the second line checks a course for appUser2
i wanted it to be easier to read - hence its split into 2 lines.

Anyways, tomorrow i will split this method into smaller methods and it should be more clear

oriAdler: Ok, now I understand your intention, and it makes sense that these two lines are separated

homepage/models.py
LeonRabinovich: @VeraTs 
I had no luck using values_list.
even though I specify ('course') as the parameter for values_list - i still get IDs instead of objects.

I decided to keep the previous code (as i sent in the picture).
This will be updated in the new PR.

**10231 and 10400 are the courses IDs I'm asserting**
image

and this is the code 
image

PR #78, opened 2021-04-27
Adding Prerequisites methods and tests

### LINKED ISSUE #77: Add Prerequisites methods and testing
At least these for now:
- return the prerequisites  for a given Course
- return if there are prerequisites for a given Course

VeraTs:
Added 2 static methods for Prerequisites model, and testing for the Prerequisites model
fixes #77

### DIFF
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -67,7 +67,15 @@ class Prerequisites(models.Model):
     req_code = models.SmallIntegerField(choices=Req_Code.choices, default=Req_Code.NONE)
 
     def __str__(self):
-        return f'req_cour = {self.req_course_id}, desired_cour = {self.course_id}, req_code = {self.req_code}'
+        return f'Req. Course = {self.req_course_id}, Desired Course = {self.course_id}, Req. Code = {self.req_code}'
+
+    @staticmethod
+    def get_prerequisites_for_course(course):
+        return Prerequisites.objects.filter(course_id=course)
+
+    @staticmethod
+    def does_course_have_prerequisites(course):
+        return Prerequisites.get_prerequisites_for_course(course).exists()
 
 
 class AppUser(models.Model):
--- a/homepage/tests.py
+++ b/homepage/tests.py
@@ -1,5 +1,5 @@
 import pytest
-from homepage.models import Course, Review
+from homepage.models import Course, Review, Prerequisites
 from django.core.exceptions import ValidationError
 from datetime import datetime
 import pytz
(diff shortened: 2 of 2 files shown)


homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+        preq.clean_fields()
+    except ValidationError:
+        flag = False
+
+    assert flag
+
+    preq.save()
+
+    assert Prerequisites.objects.filter(pk=preq.id).exists()
+    assert Prerequisites.objects.filter(pk=preq.id).count() == 1
```
ellaml: I think it's a bit redundant since the DB won't allow creation of two Preq' with the same PK.
Maybe assert Prerequisites.objects.filter(course_id=courses[0]).count() == 1 to check that only one row was created for this course would be better, or just leave the first assert.

homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+
+@pytest.mark.django_db
+# Course1 has no prerequisites
+def test_get_prerequisites_for_course_with_no_preqs(courses):
+
+    assert not Prerequisites.get_prerequisites_for_course(courses[0]).exists()
+
+
+@pytest.mark.django_db
+# test to see it retuerns True
```
ellaml: typo in comment

homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+
+
+@pytest.mark.django_db
+# test for Course2 which has prerequisite - Course1
+def test_get_prerequisites_for_course_with_one_preqs(courses):
+    preq = Prerequisites(course_id=courses[1], req_course_id=courses[0], req_code=Prerequisites.Req_Code.BEFORE)
+    preq.save()
+
+    assert Prerequisites.get_prerequisites_for_course(courses[1]).exists()
+    assert Prerequisites.get_prerequisites_for_course(courses[1]).count() == 1
```
ellaml, line 172: The first assertion is redundant since the second one will fail if it does not exist, so it covers both checks.
May I suggest retrieving the list of preq' from the method and then:
```suggestion
    preq_list = Prerequisites.get_prerequisites_for_course(courses[1])
    assert preq_list[0] == preq and len(preq_list) == 1
```
That way you're checking the full object and assure that only one was returned.

ellaml: The first assertion is redundant since the second one will fail if it does not exist, so it covers both checks.
May I suggest retrieving the list of preq' from the method and then:
```suggestion
    preq_list = Prerequisites.get_prerequisites_for_course(courses[1])
    assert preq_list[0] == preq and len(preq_list) == 1
```
That way you're checking the full object and assure that only one was returned.

homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+# test for Course2 which here depends on taking Course1 and Course3 both
+def test_get_prerequisites_for_course_with_mult_preqs(courses):
+    preqs = [Prerequisites(course_id=courses[1], req_course_id=courses[0], req_code=Prerequisites.Req_Code.BEFORE),
+             Prerequisites(course_id=courses[1], req_course_id=courses[2], req_code=Prerequisites.Req_Code.BEFORE)]
+
+    for preq in preqs:
+        preq.save()
+
+    assert Prerequisites.get_prerequisites_for_course(courses[1]).exists()
+    assert Prerequisites.get_prerequisites_for_course(courses[1]).count() == 2
```
ellaml: same as above ^

ellaml:
@VeraTs Over all looks great! I left a few comments.
Note that you have conflicts to resolve though

homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+    flag = True
+    preq = Prerequisites(course_id=courses[0], req_course_id=courses[1], req_code=Prerequisites.Req_Code.BEFORE)
+    try:
+        preq.clean_fields()
+    except ValidationError:
+        flag = False
+
+    assert flag
+
+    preq.save()
```
oriAdler: Maybe move this line into the ```try``` block, to save to the database only when valid

homepage/tests.py
```diff
@@ -103,3 +104,117 @@ def test_create_multiple_courses():
...
+
+    assert Prerequisites.objects.filter(pk=preq.id).exists()
+    assert Prerequisites.objects.filter(pk=preq.id).count() == 1
+
+
+# creating a new invalid prerequisite log - req_code invalid
+@pytest.mark.django_db
+def test_create_new_invalid_prerequisite(courses):
+    flag = False
```
oriAdler: Suggested change ```flag``` --> ```invalid```

homepage/tests.py
VeraTs: yeah, I messed this up, I intended to check for the course, will change

homepage/tests.py
VeraTs: I'll try it!

homepage/tests.py
VeraTs: done :)

oriAdler (approved):
LGTM

ellaml (approved):

homepage/models.py
```diff
@@ -69,6 +69,15 @@ class Req_Code(models.IntegerChoices):
     def __str__(self):
         return f'req_cour = {self.req_course_id}, desired_cour = {self.course_id}, req_code = {self.req_code}'
 
+    @staticmethod
+    def get_prerequisites_for_course(course):
+        preqs = Prerequisites.objects.filter(course_id=course)
+        return preqs
```
kobihk:
```suggestion
        return Prerequisites.objects.filter(course_id=course)
```

homepage/models.py
```diff
@@ -69,6 +69,15 @@ class Req_Code(models.IntegerChoices):
...
         return f'req_cour = {self.req_course_id}, desired_cour = {self.course_id}, req_code = {self.req_code}'
 
+    @staticmethod
+    def get_prerequisites_for_course(course):
+        preqs = Prerequisites.objects.filter(course_id=course)
+        return preqs
+
+    @staticmethod
+    def does_course_have_prerequisites(course):
+        return Prerequisites.objects.filter(course_id=course).exists()
```
kobihk: maybe you can use the previous function instead of repeat the same code
```suggestion
        return get_prerequisites_for_course(course).exists()
```

homepage/tests.py
```diff
@@ -191,3 +191,111 @@ def test_print_details(capsys, review_id, expected_review_details):
...
+# -----------prerequisites tests----------- #
+
+
+# three courses:
+@pytest.fixture
+def courses():
+    courses_list = [
+         Course(1, "Course1", True, 3, "N/A", 1.3, 4, 5, 1),
+         Course(2, "Course2", False, 2, "N/A", 3.5, 2, 13, 5),
+         Course(3, "Course3", False, 1, "N/A", 2, 2.5, 0, 0)]
```
kobihk: It will be more readable when we can easily see the end of the list
```suggestion
         Course(3, "Course3", False, 1, "N/A", 2, 2.5, 0, 0)
   ]
```

homepage/tests.py
```diff
@@ -191,3 +191,111 @@ def test_print_details(capsys, review_id, expected_review_details):
...
+def test_create_new_prerequisite(courses):
+    valid = True
+    preq = Prerequisites(course_id=courses[0], req_course_id=courses[1], req_code=Prerequisites.Req_Code.BEFORE)
+    try:
+        preq.clean_fields()
+        preq.save()
+    except ValidationError:
+        valid = False
+
+    assert valid and Prerequisites.objects.filter(course_id=preq.course_id).count() == 1
```
kobihk, line 229: I think you can reuse the method get_prerequisites_for_course
```suggestion
    assert valid and Prerequisites.get_prerequisites_for_course(preq.course_id).count() == 1
```

homepage/models.py
```diff
@@ -69,6 +69,15 @@ class Req_Code(models.IntegerChoices):
     def __str__(self):
         return f'req_cour = {self.req_course_id}, desired_cour = {self.course_id}, req_code = {self.req_code}'
```
kobihk: I know this is not part of this PR but because you test it in line # 243 I pay attention to this string.
IMHO it will be more readable like:
```suggestion
        return f'Req course id: {self.req_course_id}, Desired course id: {self.course_id}, Req code: {self.req_code}'
```

homepage/tests.py
```diff
@@ -191,3 +191,111 @@ def test_print_details(capsys, review_id, expected_review_details):
...
+def test_does_course_have_prerequisites_one_preqs(courses):
+    preq = Prerequisites(course_id=courses[1], req_course_id=courses[0], req_code=Prerequisites.Req_Code.BEFORE)
+    preq.save()
+
+    assert Prerequisites.does_course_have_prerequisites(courses[1])
+
+
+@pytest.mark.django_db
+# test to see it returns True
+def test_does_course_have_prerequisites_for_mult_preqs(courses):
```
kobihk:
```suggestion
def test_does_course_have_prerequisites_for_multi_preqs(courses):
```

homepage/tests.py
```diff
@@ -191,3 +191,111 @@ def test_print_details(capsys, review_id, expected_review_details):
...
+    preqs = [Prerequisites(course_id=courses[1], req_course_id=courses[0], req_code=Prerequisites.Req_Code.BEFORE),
```
kobihk: I can see many tests that have almost the same code. 
did you try to use parameterized?

kobihk (changes requested):

homepage/tests.py
VeraTs: @kobihk but it's before I test the function itself, this is a more basic test so it should come first, no?

homepage/tests.py
VeraTs: As I don't want it to create the preqs sometimes, I put it inside a fixture

homepage/tests.py
kobihk: I think it should be OK because it's only calling to a method that doing the same as you have here. 
but if you want to leave it like that it's ok by me :)

homepage/tests.py
```diff
@@ -191,3 +191,112 @@ def test_print_details(capsys, review_id, expected_review_details):
...
+def test_does_course_have_prerequisites_one_preqs(courses, preqs_list):
+    preq = preqs_list[0]
+    preq.save()
+
+    assert Prerequisites.does_course_have_prerequisites(courses[1])
+
+
+@pytest.mark.django_db
+# test to see it returns True
+def test_does_course_have_prerequisites_for_mult_preqs(courses, preqs_list):
```
kobihk:
```suggestion
def test_does_course_have_prerequisites_for_multi_preqs(courses, preqs_list):

```

kobihk (changes requested):

kobihk (approved):

PR #80, opened 2021-04-27
Adding test and view methods for Course model

### LINKED ISSUE #79: Add Course methods for views
Methods that display all or some of the courses, as follows:
- get_all_courses
- Sorting: to get a list or a QuerySet of Course objects, sorted in a given manner (id/name/rating)
- Filtering: to get a list or QuerySet of Course objects that answer a specific term, such as avg_rating bigger than a given value

VeraTs:
This is based on PR #78 
Fixes #79 

contains:
- template for courses page with filter form
- filtering form - multiple choice - checkbox widget
- new methods for Course model - intended for use in view and shell
- tests for new methods and for view in courses_view_tests.py

clean page:
image
picking out a filter:
image
after filtering:
image

### DIFF
--- a/homepage/forms.py
+++ b/homepage/forms.py
@@ -12,3 +12,13 @@ class ReviewForm(ModelForm):
             'rate': forms.Select(choices=CHOICES),
             'course_load': forms.Select(choices=CHOICES),
         }
+
+
+class FilterForm(forms.Form):
+    choices = [('rate_over', 'rating over 3.5'), ('load_below', 'course load under 3.5'),
+               ('mand', 'mandatory'), ('elect', 'elective')]
+    filter_by = forms.MultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple,
+        choices=choices,
+        required=False
+        )
--- a/homepage/models.py
+++ b/homepage/models.py
@@ -26,7 +26,7 @@ class Course(models.Model):
         return self.name
 
     def print_details(self):
-        mandatory = 'yes' if self.mandatory == 1 else 'no'
+        mandatory = 'yes' if self.mandatory else 'no'
         msg = (
             "------------------------------------------------------------\n"
             f"Course indentifier: {self.course_id}   \nName: {self.name}\nMandatory? {mandatory}\n"
@@ -39,6 +39,36 @@ class Course(models.Model):
     def get_details(self):
         return (self.course_id, self.name, self.mandatory, self.credit_points, self.syllabi,
                 self.avg_load, self.avg_rating, self.num_of_raters, self.num_of_reviewers)
+
+    # --- returns all Course objects - the main 'courses' source for the view
+    @staticmethod
+    def get_courses():
+        return Course.objects.all()
+
+    # --- get filtered course list by rating/load/mandatory/elective  - return QuerySets
+    @staticmethod
+    def get_filtered_courses_by_rating(rating, courses):
+        # gets all courses with average rating >= rating
+        return courses.filter(avg_rating__gte=rating)
+
+    @staticmethod
+    def get_filtered_courses_by_load(load, courses):
+        # gets all courses with average course load <= load
+        return courses.filter(avg_load__lte=load)
+
+    @staticmethod
+    def get_mandatory_courses(courses):
+        # gets all courses that are mandatory
+        return courses.filter(mandatory=True)
+
+    @staticmethod
+    def get_elective_courses(courses):
+        # gets all courses that are electives
+        return courses.filter(mandatory=False)
+
+    # --- returns if course has Prerequisites
+    def has_preqs(self):
+        return Prerequisites.does_course_have_prerequisites(self)
 
 
 class Prerequisites(models.Model):
--- a/homepage/tests/courses_view_tests.py
+++ b/homepage/tests/courses_view_tests.py
@@ -1,0 +1,187 @@
+import pytest
+from homepage.models import Course, Prerequisites
+from homepage.forms import FilterForm
+from pytest_django.asserts import assertTemplateUsed
+
+
+# --- testing the filtering functions:
+@pytest.fixture
+def courses():
+    courses_list = [
+         Course(1, "Course1", True, 3, "N/A", 1.3, 4, 5, 1),
+         Course(2, "Course2", False, 2, "N/A", 3.5, 2, 13, 5),
+         Course(3, "Course3", False, 1, "N/A", 2, 2.5, 0, 0)
+         ]
+    for course in courses_list:
+        course.save()
+
+    return courses_list
+
+
+@pytest.fixture
+def preqs_list(courses):
+    preqs = [Prerequisites(course_id=courses[1], req_course_id=courses[0], req_code=Prerequisites.Req_Code.BEFORE),
+             Prerequisites(course_id=courses[1], req_course_id=courses[2], req_code=Prerequisites.Req_Code.BEFORE)]
+    for preq in preqs:
+        preq.save()
+    return preqs
+
+
+@pytest.fixture
+def all_courses():
+    return Course.objects.all()
+
+
+# test that each of the new courses is returned with the method
+@pytest.mark.django_db
+def test_get_courses(courses):
+    flag = True
+    db_courses = Course.get_courses()
+    for course in courses:
+        flag = flag and db_courses.filter(pk=course.pk).exists()
+
+    assert flag
+
+
+# test the has_preqs method - if a course with preqs returns True and withuot returns False
+@pytest.mark.django_db
+def test_course_has_preqs(courses, preqs_list):
+    has_preqs = []
+    desired_res = [False, True, False]
+    for course in courses:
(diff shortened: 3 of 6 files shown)


homepage/models.py
```diff
@@ -40,6 +40,65 @@ def get_details(self):
         return (self.course_id, self.name, self.mandatory, self.credit_points, self.syllabi,
                 self.avg_load, self.avg_rating, self.num_of_raters, self.num_of_reviewers)
 
+    # --- get all courses sorted by id/name/rating  - return QuerySets
+    @staticmethod
+    def get_all_courses_sorted_by_id():
+        return Course.objects.all()
```
oriAdler, line 46: Missing here ```.order_by```, or is ordered by id by default?
Maybe worth a comment

homepage/models.py
```diff
@@ -40,6 +40,65 @@ def get_details(self):
...
+
+    @staticmethod
+    def get_all_courses_sorted_by_name():
+        courses = Course.objects.order_by('name')
+        return courses
+
+    @staticmethod
+    def get_all_courses_sorted_by_rating():
+        courses = Course.objects.order_by('avg_rating')
```
oriAdler:
```suggestion
        return Course.objects.order_by('-avg_rating')  # descending order

```

homepage/models.py
```diff
@@ -40,6 +40,65 @@ def get_details(self):
...
+    def course_list_sort_by_rating(course_list):
+        # gets a list of course object instances and sorts it by rating
+        course_list.sort(key=lambda course: course.rating)
+        return course_list[::-1]
+
+    # --- get filtered course list by rating/load/mandatory/elective  - return QuerySets
+    @staticmethod
+    def get_filtered_courses_by_rating(rating):
+        # gets all courses with average rating >= rating
+        courses = Course.objects.filter(avg_rating__gte=rating)
```
oriAdler:
```suggestion
        return Course.objects.filter(avg_rating__gte=rating)

```

homepage/models.py
```diff
@@ -40,6 +40,65 @@ def get_details(self):
...
+        return courses
+
+    @staticmethod
+    def get_mandatory_courses():
+        # gets all courses that are mandatory
+        courses = Course.objects.filter(mandatory=True)
+        return courses
+
+    @staticmethod
+    def get_elective_courses():
```
oriAdler: Not a big difference, but maybe you can combine this method with ```get_mandatory_courses```,
and pass ```mandatory``` as parameter

oriAdler (commented):
In general, it looks Good!
I'm not sure which approach is preferred. 
Another consideration, which approach is easier to wire to the UI?
Notice that on the Challenge, Barak decided to return a query set to the UI.

homepage/models.py
VeraTs: ordered by id as default, yes

homepage/models.py
VeraTs: will change!

homepage/models.py
VeraTs: will change

homepage/models.py
VeraTs: I like the separation, myself, because I can't think of a name other than "get_mandatory_yay_nay" to describe a method that does both depending on input

VeraTs:
@oriAdler I tried altering the sorting functions and now they work with QuerySet too! So the dilemma was resolved

oriAdler:
@VeraTs Glad to hear that, so we can enjoy the best of both worlds

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,180 @@
...
+
+@pytest.mark.django_db
+def test_get_elective_courses(courses, all_courses):
+    # should get back only Course1 from courses
+    db_courses = Course.get_elective_courses(all_courses)
+
+    assert not db_courses.filter(pk=courses[0].course_id)
+    assert db_courses.filter(pk=courses[1].course_id) and db_courses.filter(pk=courses[2].course_id)
+
+# -- testing double filter
```
ellaml, line 94: ?

VeraTs: like why am I testing a filter over a filter?
I wanted to make sure a user can choose multiple filters and receive a proper result
It may be redundant, I'm not sure yet.

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,180 @@
...
+    courses = list(Course.get_courses().values_list('course_id'))
+    response = client.get('/courses/')
+    assert response.status_code == 200
+    view_courses = list(response.context['all_courses'].values_list('course_id'))
+    assert courses == view_courses
+
+
+# test that the post response is the desired form model
+@pytest.mark.django_db
+def test_post_form(client):
```
oriAdler:
```suggestion
def test_uses_filter_form(client):
```

oriAdler (commented):
Looks great!
I would suggest adding the ```assertTemplateUsed``` test like in #84.

oriAdler (approved):
LGTM

homepage/forms.py
```diff
@@ -0,0 +1,10 @@
+from django import forms
+
+
+class FilterForm(forms.Form):
+    choices = [('rate_over', 'rating over 3.5'), ('load_below', 'course load under 3.5'),
+               ('mand', 'mandatory'), ('elect', 'elective')]
+    filter_by = forms.MultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple,
+        choices=choices,
+        required=False)
```
kobihk: it's more readable when the close parenthesis is in a new line
```suggestion
        required=False
   )
```

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,187 @@
...
+    assert db_courses.filter(pk=courses[2].course_id)
+    assert not db_courses.filter(pk=courses[1].course_id) and not db_courses.filter(pk=courses[0].course_id)
+
+
+# test mandatory and elective filter - expect empty result
+@pytest.mark.django_db
+def test_elective_and_mandatory_filters(all_courses):
+    # expect to return only Course1
+    db_courses = Course.get_elective_courses(all_courses)
+    db_courses = Course.get_mandatory_courses(db_courses)
```
kobihk, line 124: same ^^

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,187 @@
...
+
+# -- testing double filter
+
+
+# test rating and mandatory filter
+@pytest.mark.django_db
+def test_rating_and_mandatory_filters(courses, all_courses):
+    # expect to return only Course1
+    db_courses = Course.get_filtered_courses_by_rating(2.5, all_courses)
+    db_courses = Course.get_mandatory_courses(db_courses)
```
kobihk, line 102: looks like you overwrite the db_courses from the previous line

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,187 @@
...
+    assert db_courses.filter(pk=courses[0].course_id)
+    assert not db_courses.filter(pk=courses[1].course_id) and not db_courses.filter(pk=courses[2].course_id)
+
+
+# test rating and elective filter
+@pytest.mark.django_db
+def test_rating_and_elective_filters(courses, all_courses):
+    # expect to return only Course3
+    db_courses = Course.get_filtered_courses_by_rating(2.5, all_courses)
+    db_courses = Course.get_elective_courses(db_courses)
```
kobihk, line 113: same ^^

homepage/tests/courses_view_tests.py
```diff
@@ -0,0 +1,187 @@
...
+    assert db_courses_mand.filter(pk=courses[0].course_id)
+    assert not db_courses_mand.filter(pk=courses[1].course_id) and not db_courses_mand.filter(pk=courses[2].course_id)
+
+
+# test rating and load
+@pytest.mark.django_db
+def test_get_rating_and_load_filters(courses, all_courses):
+    # expect only Course1
+    db_courses = Course.get_filtered_courses_by_load(2, all_courses)
+    db_courses = Course.get_filtered_courses_by_rating(3, db_courses)
```
kobihk, line 151: same ^^

kobihk (changes requested):

homepage/tests/courses_view_tests.py
ellaml: No it's not redundant, sorry for the non informative comment it was a long day
I just didn't understand which section the comment is referring to, thought you left it there by mistake

homepage/tests/courses_view_tests.py
VeraTs: I did overwrite it - because I wanted to simulate the way it would work in application:
you have an initial QuerySet that is filtered once, and put the result into the same variable
and then it is filtered again, also being put into the same variable

kobihk: I see. 
Thanks for the detailed explanation!!

kobihk (approved):

PR #82, opened 2021-04-28
Add post review feature

### LINKED ISSUE #81: Add post review feature
1) Add a form to the ```add_review``` page to enable posting a review
2) Add tests to check the new feature

oriAdler:
1) Add **forms.py** to create a ```ReviewForm```
**Note:** post image/histogram feature will be add after the third demo
2) Modify **add_review.html** to render the review form
3) Add **post_review_tests.py** to test the new feature
4) Modify **routing_tests.py** - add ```@pytest.mark.django_db```
because now **add_review** page demands access to the database
5) Modify ```def add_review(request)``` in **views.py** to enable **Post**
and render the template

fix: #81

### DIFF
--- a/homepage/forms.py
+++ b/homepage/forms.py
@@ -1,0 +1,14 @@
+from django import forms
+from django.forms import ModelForm
+from .models import Review
+
+
+class ReviewForm(ModelForm):
+    class Meta:
+        model = Review
+        fields = ['course', 'user', 'rate', 'content', 'course_load', 'professor']
+        CHOICES = [['', '---'], [1, '1'], [2, '2'], [3, '3'], [4, '4'], [5, '5']]
+        widgets = {
+            'rate': forms.Select(choices=CHOICES),
+            'course_load': forms.Select(choices=CHOICES),
+        }
--- a/homepage/tests/post_review_tests.py
+++ b/homepage/tests/post_review_tests.py
@@ -1,0 +1,66 @@
+import pytest
+from homepage.forms import ReviewForm
+from homepage.models import Review
+from pytest_django.asserts import assertTemplateUsed
+
+
+@pytest.fixture
+def review_details():
+    return {'course': 10231, 'user': 3, 'rate': 5, 'content': "Great course!", 'course_load': 4, 'Professor': ''}
+
+
+# --------Backend testing-------- #
+@pytest.mark.parametrize("invalid_review_details", [
+    {'course': 10231, 'user': 3, 'rate': 10, 'content': "Great course!", 'course_load': 4, 'Professor': None},
+    # review_details0:                   ^rate > 5
+    {'course': 10231, 'user': 3, 'rate': 4, 'content': "Great course!", 'course_load': -2, 'Professor': None},
+    # review_details1:                                                                 ^course_load < 0
+    {'course': 10231, 'user': '', 'rate': 4, 'content': "Great course!", 'course_load': -2, 'Professor': None},
+    # review_details2:        ^user field is required
+])
+@pytest.mark.django_db
+def test_post_invalid_review(invalid_review_details):
+    invalid = False
+    form = ReviewForm(data=invalid_review_details)
+
+    try:
+        form.save()
+    except ValueError:
+        invalid = True
+
+    assert invalid
+
+
+@pytest.mark.django_db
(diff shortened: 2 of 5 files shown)


homepage/forms.py
```diff
@@ -0,0 +1,14 @@
+from django import forms
+from django.forms import ModelForm
+from .models import Review
+
+
+class ReviewForm(ModelForm):
+    class Meta:
+        model = Review
+        fields = ['course', 'user', 'rate', 'content', 'course_load', 'professor']
+        CHOICES = [['', '---'], [1, '1'], [2, '2'], [3, '3'], [4, '4'], [5, '5']]
```
ellaml, line 10: Why ['','---'] is one of the options? Both fields are mandatory
Is the first option the one that is displayed as default?
I'm not sure how the choices dropdown works yet but I think we should avoid allowing the user to leave them empty,
even if that means that by default the rating will be 3 or something

homepage/views.py
```diff
@@ -18,4 +19,11 @@ def reviews(request):
 
 
 def add_review(request):
-    return render(request, 'homepage/add_review.html')
+    if request.method == "POST":
+        form = ReviewForm(request.POST)
+        if form.is_valid():
+            form.save()
+            return redirect('add_review')
```
ellaml, line 26: What happens if it's invalid? We stay on the same page and nothing happens until the user sends valid data?
I also saw you've written a test for valid data with client to return "302" code, but haven't written one for invalid, so I'm not sure what's the logic in that case. Even if nothing happens maybe we should test that or at least add some comments about it

homepage/forms.py
VeraTs: Where are the other 'preset' fields for the dorm, such as getting a course from the approved list and such? are we leaving them for after the demo?

oriAdler: image
@ellaml The ```''``` makes the user to chose an option, the ```'---'``` is what rendered, I tried to be consistent with the other fields. So the user must choose a number to post a review.

oriAdler: image
@VeraTs  Django gives us a gift since we made the connection between our models, by default the form shows options only from the approved list.

homepage/views.py
oriAdler: If it's invalid when the user push the ```post``` button, a popup window will be shown (It's Django built-in feature)
So I wasn't sure how to test with ```client``` the invalid option, as the popup window is coupled.
We do have a test for it from the Backend.
Maybe a test with the client that tries to post invalid data, and check if the QuerySet size stayed the same before and after, WDYT?

ellaml: In my opinion the way from submitting the form till retrieving back the updated list of object is too long to be the one that determines whether a validation check was applied. Other things can go wrong on the way so passing the test does not necessarily indicates the this Django feature was taking place.
I think it's ok to leave it like that and trust Django, but let's see what the mentors think.

homepage/forms.py
ellaml: great!

ellaml (approved):
LGTM

homepage/forms.py
VeraTs: awesome :)

VeraTs (approved):

homepage/tests/post_review_tests.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+# --------Backend testing-------- #
+@pytest.mark.parametrize("invalid_review_details", [
+    {'course': 10231, 'user': 3, 'rate': 10, 'content': "Great course!", 'course_load': 4, 'Professor': None},
+    # review_details0:                   ^rate > 5
+    {'course': 10231, 'user': 3, 'rate': 4, 'content': "Great course!", 'course_load': -2, 'Professor': None},
+    # review_details1:                                                                 ^course_load < 0
+    {'course': 10231, 'user': '', 'rate': 4, 'content': "Great course!", 'course_load': -2, 'Professor': None},
+    # review_details2:        ^field is required
```
kobihk:
```suggestion
    # review_details2:        ^user field is required
```

kobihk (changes requested):

Yarboa (approved):
LGTM

kobihk (approved):

PR #84, opened 2021-04-29
Adding Reviews Page and static folder

### LINKED ISSUE #83: Add Reviews page
Add a page that displays all the reviews in the app (for all courses), ordered by date

ellaml:
* Modified `reviews.html `template to display all the reviews
* Added Static folder and added the STATIC_URL from settings file to `urls.py` in order to
  load it in templates
* Added "css" folder and icons folder in "static" folder
* Added `@pytest.mark.django_db` to `routing_tests.py` since it uses the
  objects from the db
* Added `reviews_tests.py` file with one test to check that the template
  was loaded (no other logic was applied in the page yet)
* Modified 'review_test_data' file that each Review has a different date to test the sorting
* Added main_feed() method to return the sorted data and a test for it
fix: #83 

image


ellaml:
Note that I moved the { % extends ... % } & { % block content % } (app_layout) to the beginning of the html file, since is loads the bootstrap link and the CSS link was not applied unless it appeared after the bootstrap link.
So I figured out it's better than link the CSS file in the` <body>` instead of `<head>`

homepage/views.py
```diff
@@ -14,7 +15,8 @@ def courses(request):
 
 
 def reviews(request):
-    return render(request, 'homepage/reviews/reviews.html')
+    reviews = Review.objects.order_by('-date')
```
oriAdler:
```suggestion
    reviews = Review.main_feed()
```
Write a function ```def main_feed(cls)``` like in **GUI-TALK-v2** page 29
Finally, you can write a similar test:
image

homepage/templates/homepage/reviews/reviews.html
```diff
@@ -1,18 +1,77 @@
...
+              <img src="{% static 'icons/star.png' %}">
+              {% endfor %}
+              {% else %}
+              <img src="{% static 'icons/star.png' %}">
+              {% endif %}
+        </div>
+        <p style="margin-top: 20px">
+              {{review.content}}
+              <div class="extra-details">
+               {% if review.professor %}
```
oriAdler: I've noticed that when the professor's name is in the review, the **course_load** appears on the right instead of below.
Is it possible to change that?
image

oriAdler (commented):
image
The page looks great!
I left a few comments.

homepage/templates/homepage/reviews/reviews.html
ellaml: I think it won't look good with nothing on the left side

homepage/views.py
ellaml: @oriAdler gonna change the test data to hold different date in each review (cause otherwise the sort has no meaning), and add this function and test for it

homepage/templates/homepage/reviews/reviews.html
VeraTs: I like it when it's on the right - maybe put it there permanently regardless of prof?

VeraTs (commented):
The screenshot looks amazing, wonderful application of icons!

oriAdler (approved):
LGTM

homepage/templates/homepage/reviews/reviews.html
```diff
@@ -1,18 +1,77 @@
...
+    <div class="row likes-row">
+         <img class="like-btn" src="{% static 'icons/thumb-up.png' %}"> {{review.likes_num}} students found this helpful
+   </div>
+</p>
 </div>
-<p class="EOM">(End of messages)</p>
-{% endblock %}
+{% endfor %}
+</div>
+</div>
```
kobihk: these tags for example lose their indentation 
please try to reformat this page to get it better indentation
In PyCharm is: CTR+ALT+L

kobihk (changes requested):

ellaml:
@oriAdler @VeraTs moved **course load** to the right

VeraTs (approved):

kobihk (approved):

PR #96, opened 2021-05-05
Add Basic Sign-in, Sign-up Pages

### LINKED ISSUE #88: Add Basic Signup, Signin Pages

LeonRabinovich:
### Edit: as requested, this was split into 2 different PRs (1 for each page)
___________________________________________________________________________________
- Installed django-crispy-forms (used in the HTML pages)
- Sign-in and Sign-out pages are available at 127.0.0.1:8000/**users/sign_in/**, 127.0.0.1:8000/**users/sign_up/**
- Sign-in, Sign-up use Django built-in validations for username, password, password confirmation 
- If Sign-up is successful, page is redirected to Sign-in page (over wise the form is cleared)
- If Sign-in is successful, page is redirected to homepage (over wise the form is cleared)
- More features will be implemented in a separate PR 
( Sign-in and Sign-up success/fail messages, email validation for mta@ac.il, etc..)

Resolves: #88


oriAdler:
@LeonRabinovich Maybe it's better to separate this PR into two PR's Sign-in page and Sign-up page? (if it's not too much mess to separate)
It will be easier to make a CR.
Also, it can be approved faster.
WDYT?

LeonRabinovich:
> @LeonRabinovich Maybe it's better to separate this PR into two PR's Sign-in page and Sign-up page? (if it's not too much mess to separate)
> It will be easier to make a CR.
> Also, it can be approved faster.
> WDYT?

Well it's possible but it will cause some problems because they Are tightly connected together at some methods.(I will probably need to remove some stuff)

Also, i can only work on it next week, and it will prolong the next PR for the user profile page.

If you think it's critical to split them - let me know.

LeonRabinovich:
> @LeonRabinovich Maybe it's better to separate this PR into two PR's Sign-in page and Sign-up page? (if it's not too much mess to separate)
> It will be easier to make a CR.
> Also, it can be approved faster.
> WDYT?

I'll split it into 2 diffrent PRs, but probably only on Friday / Saturday

PR #99, opened 2021-05-07
Add basic sign-in page

### LINKED ISSUE #88: Add Basic Signup, Signin Pages

LeonRabinovich:
Resolves: #88 

- Installed django-crispy-forms (used in the HTML pages)
- Sign-in page is available at 127.0.0.1:8000/**users/sign_in/** (And from navigation bar)
- If Sign-in is successful, page is redirected to homepage (over-wise the form is cleared)
- Implemented fail/success messages  
- Replaced sign-in button with sign-out button after successful sign-in
- More features will be implemented in a separate PR

**sign-in, sign-up, sign-out demonstration video:**

https://user-images.githubusercontent.com/79373827/118169108-3b868180-b431-11eb-8261-ac304659dd6e.mp4


homepage/views.py
```diff
@@ -53,3 +56,26 @@ def add_review(request):
...
+            if user is not None:
+                login(request, user)
+                messages.info(request, f"You are now logged in as {username}.")
+                return redirect("landing")
+            else:
+                messages.error(request, "Invalid username or password.")
+
+        else:
+            messages.error(request, "Invalid username or password.")
+
```
VeraTs, line 106: I see you've used the messages framework from django, but is it displayed anywhere?
If I understand your structure correctly, then there is supposed to be a display of this error message at the marked line in the sign_in html file.

homepage/templates/homepage/users/sign_in.html
```diff
@@ -0,0 +1,17 @@
+{% extends "homepage/app_layout.html" %}
+
+{% block content %}      
+
+{% load crispy_forms_tags %}
+
+<div class="container py-5">
+    <h1>Sign In</h1>
+    <form method="POST">
```
VeraTs, line 9: Error message display is supposed to be here, no?

VeraTs (commented):
Great work
Good call on the crispy form addition, from what I understand it will look way better after some manipulation with it.
My concern - you have error messages, as you should, but I saw no display of them, as noted here:

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        user = authenticate(username=not_signed_up_details[0], password=not_signed_up_details[1])
+        assert user is None
+
+    def test_signin_valid(self, signed_up_details):
+        user = authenticate(username=signed_up_details[0], password=signed_up_details[1])
+        assert user
+
+    # ------------------------Front-End testing------------------------ #
+
+    def test_post_valid_signIn_with_client(self, client, signed_up_details):
```
oriAdler: I would suggest adding another test to check invalid sign-in with the client.
To make sure the website redirects to the same page.

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_signin_invalid(self, not_signed_up_details):
+        user = authenticate(username=not_signed_up_details[0], password=not_signed_up_details[1])
+        assert user is None
+
+    def test_signin_valid(self, signed_up_details):
+        user = authenticate(username=signed_up_details[0], password=signed_up_details[1])
+        assert user
+
+    # ------------------------Front-End testing------------------------ #
```
oriAdler, line 30: You can also test that the new page is using ```AuthenticationForm```, take a lot on PR #82

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_post_valid_signIn_with_client(self, client, signed_up_details):
+        response = client.post('/users/sign_in/', data={
+            'username': signed_up_details[0], 'password': signed_up_details[1]})
+
+        assert response.status_code == 302
```
oriAdler, line 37: Maybe you can check that after authentication, the user is redirected to the landing page.
Like in the presentation:
image

oriAdler (commented):
Looks great!
Good catch with the crispy forms.
I've left a few comments

homepage/views.py
LeonRabinovich: Error messages are only partly implemented and will be fully implemented on a separate PR with other features
I've added a comment in the PR description for more clarity.

homepage/templates/homepage/users/sign_in.html
LeonRabinovich: Error messages are only partly implemented and will be fully implemented on a separate PR with other features
I've added a comment in the PR description for more clarity.

homepage/tests/sign_in_tests.py
LeonRabinovich: Added

homepage/tests/sign_in_tests.py
LeonRabinovich: Added a check to test redirection to homepage after successful sign-in

homepage/tests/sign_in_tests.py
LeonRabinovich: added a check to test invalid sign-in redirection to the sign-in page

oriAdler (approved):
LGTM

homepage/views.py
VeraTs: As I've said in the PR #100 - this is implemented enough to display, so please display.

VeraTs:
@LeonRabinovich as in PR #100 please un-resolve the conversation above, and continue.

oriAdler (approved):

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_signin_valid(self, signed_up_details):
+        user = authenticate(username=signed_up_details[0], password=signed_up_details[1])
+        assert user
+
+    # ------------------------Front-End testing------------------------ #
+
+    def test_post_valid_signIn_with_client(self, client, signed_up_details):
+        response = client.post('/users/sign_in/', data={
+            'username': signed_up_details[0], 'password': signed_up_details[1]})
+
```
kobihk: please remove the blank line.
I would use a blank line as a "separator" if the method is too long and I have too many cases to separate them but in cases of a small method I think we don't have to add it.

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,56 @@
...
+        HOMEPAGE_URL = '/'
+        assert response.url == HOMEPAGE_URL
+        assert response.status_code == 302
+
+    def test_invalid_signIn_redirect(self, client, not_signed_up_details):
+        response = client.post('/users/sign_in/', data={
+            'username': not_signed_up_details[0], 'password': not_signed_up_details[1]})
+
```
kobihk: same ^^

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,56 @@
...
+        assert response.url == '/users/sign_in/'
+
+    def test_renders_add_signIn_template(self, client):
+        response = client.get('/users/sign_in/')
+
```
kobihk: same ^^

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,56 @@
...
+        assert response.status_code == 200
+        assertTemplateUsed(response, 'homepage/users/sign_in.html')
+
+    def test_uses_authenticate_form(self, client):
+        response = client.get('/users/sign_in/')
+
```
kobihk: same ^^

homepage/views.py
```diff
@@ -53,3 +56,29 @@ def add_review(request):
...
+        if form.is_valid():
+            username = form.cleaned_data.get('username')
+            password = form.cleaned_data.get('password')
+            user = authenticate(username=username, password=password)
+
+            if user is not None:
+                login(request, user)
+                signin_sucesfull = True
+
```
kobihk: same ^^

kobihk (changes requested):

homepage/tests/sign_in_tests.py
LeonRabinovich: Removed

homepage/tests/sign_in_tests.py
LeonRabinovich: Removed

homepage/tests/sign_in_tests.py
LeonRabinovich: Removed

homepage/tests/sign_in_tests.py
LeonRabinovich: Removed

homepage/views.py
LeonRabinovich: Removed

homepage/views.py
LeonRabinovich: @VeraTs Added implementation for messages in the HTML pages

LeonRabinovich:
@VeraTs @kobihk 
In addition to implementing the messages in the HTML pages and removing blank lines from the methods,
I've also added a sign-out button (last commit) which will appear when a user is signed in.

LeonRabinovich:
@kobihk conflicts resolved

Yarboa (approved):
LGTM

homepage/templates/homepage/app_layout.html
```diff
@@ -48,6 +54,24 @@
...
 </nav>
 
+
+<style>
+  .messages {
+    background-color:dodgerblue;
+    color: white;
+    padding: 10px;
+  }
+  </style>
```
VeraTs: `<style>` should live in between the `<head>` tags

homepage/templates/homepage/app_layout.html
```diff
@@ -48,6 +54,24 @@
...
+  
+  {% if messages %}
+  <ul class="messages">
+      {% for message in messages %}
+      <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+      {% endfor %}
+  </ul>
+  {% endif %}
```
VeraTs: 66 - 72:
when I talked about displaying error messages that are raised from the form (sing up, log in) I meant that they appear right above the form itself, as marked in the related .html file below.

homepage/templates/homepage/users/sign_in.html
VeraTs, line 9: IMHO here is a good place to display errors if they are directly resultant from posting the form,
since no other place currently raises error messages to display to the user

VeraTs (commented):
I'm happy you saw my point :)
Just a few things, as written below:

homepage/templates/homepage/app_layout.html
VeraTs: Also, I think it might be easier to use the bootstrap built in component Alert for this:
```
<div class="alert alert-warning alert-dismissible fade show" role="alert">
    ...
</div>
```

LeonRabinovich: @VeraTs 
Thanks for your detailed feedback.

There are a coupe of reasons i implemented the messages display in **app_layout.html**:
(which makes **the messages appear in both the sign-in and the signup form**)

1.  **Duplicate code:** since both the sign-in and sign-up pages need to use the exact same code to display the messages, it only makes sense to implement the code once in app_layout which they both extend, instead of writing the same code twice in 2 different places. (sign_in.html and sign_up.html)

2. **The messages need to appear in other pages as well** and not only in sign_in.html and sign_up.html.
This is because after you complete the sign-in \ sign-up process - you are redirected to a different page.
Hence - implementing lines 66-72 where you suggested (sign_in.html) will not work for messages that need to be redirected to other pages.
If implemented once in app_layout.html -  all the messages are displayed in the correct page even if redirected.



- **Regarding bootstrap alerts:**
This will require some extra work because the code to display the alerts is a little different.
The only change this will make **is to add more colors to the messages** (each message type has its own color):
https://www.ordinarycoders.com/blog/article/django-messages-framework


Currently all the success / error / info messages are displayed in a blue color. (And IMO it looks good)
Ofc, in my feature PR for the sign-in and sign-up pages i might add the bootstrap alerts for more colors, **but IMO this is not a high priority and there are more important features we should focus on first** 
(we only have 10 more days for the demo...)


**Example of the current messages:**
image


image

homepage/templates/homepage/users/sign_in.html
LeonRabinovich: See my reply above

homepage/templates/homepage/app_layout.html
LeonRabinovich: Added `<head>` tags

homepage/templates/homepage/app_layout.html
oriAdler: @LeonRabinovich I agree with Vera. These messages shouldn't be in **app_layout.html**, as they have nothing in common with other features.
On all websites, when the user logs in, the user just gets redirected to the desired page or landing page.
When the user logs out, the user gets redirected to the login page.
For example, the error message ```invalid user name``` should be right below the sign-in button, and not placed somewhere on the screen.

VeraTs: @LeonRabinovich If someone tries to sign up or log in and is unsuccessful, are they redirected to a page that isn't just the same?
Since in my experience, in most websites, they are redirected (or don't even leave) the page for the sign up/ log in.

homepage/templates/homepage/app_layout.html
VeraTs: @LeonRabinovich , `<head>` tags only appear once per html page, above the `<body>` tag.
See [here](https://www.w3schools.com/html/html_intro.asp) for further reference

homepage/templates/homepage/app_layout.html
LeonRabinovich: @oriAdler 

The messages are in app_layout.html because many of our pages need to use the messages, and all pages extend app_layout.html. **Also in the future when you implement messages in view.py for any page - they will be displayed automatically because of the code inside app_layout.html**

**it makes no sense to duplicate the code and place it in 3-4 diffrent pages**. (again this is the exact same code)
This would mean if you have 4 HTML pages with messages - you will need to maintain **the same code** 4 times in 4 different places.


`On all websites, when the user logs in, the user just gets redirected to the desired page or landing page.
When the user logs out, the user gets redirected to the login page.`
A message should be displayed after a user logs-in \ logs-out (even at a different page),
this is not possible if you put the messages code only in sign_in.html.


`For example, the error message invalid user name should be right below the sign-in button, and not placed somewhere on the screen.`

This is not implemented, currently implementation is only for basic general error messages

LeonRabinovich: @VeraTs 

To make it clear, i will list all of  the scenarios for sign-in:


1. **Sign-in is successful:** redirect to the homepage. successful sign-in message is displayed after re-direction in the homepage. (this is possible because the homage extends app_layout.html)
 
2. **Sign-in fails:** stay on the same page and clear the form (or just re-direct to the sign-in page) and display sign-in failed message.

3. **Sign-out:** Redirect to the homepage and display a message (" 'UserName' signed out successfully ").
(this is possible because the homage extends app_layout.html)

- **Also : if sign-up is successful** - Redirect to the sign-in page and display a message ( 'Username' signed-up successful )
This is also possible because both pages extend app_layout.html

LeonRabinovich: @VeraTs @oriAdler 

In addition to the comments above, i have 1 more thing to add.
The purpose of this PR is to implement the sign-in and sign-out functionality in the simplest way possible with simple clear messages.

There are other ways to implement the messages, however currently I'm using the simplest solution (which works and has all the functionality we need IMO).

If we had more time (again the final demo is in 10 days), i could make the messages look better and displayed better, but please note we are on a tight schedule.
Does it matter that much if the error message is displayed in the button of the screen or the top of the screen? (for the above purpose)

**I think its better to have a simple solution that works so we can work on other features that rely on the sign-in functionality**.(for example the profile page)
If we have time - we can make make the UI look better, but i think this should not be the main goal right now.

oriAdler: @LeonRabinovich As all pages inherit from the app layout, this page is supposed to be minimal as possible.
All other pages don't use these messages.
So I believe it's best if the messages section will be on the sign-in page.
About the schedule, this feature was supposed to be completed a long time ago.

LeonRabinovich: > @LeonRabinovich As all pages inherit from the app layout, this page is supposed to be minimal as possible.
> All other pages don't use these messages.
> So I believe it's best if the messages section will be on the sign-in page.
> About the schedule, this feature was supposed to be completed a long time ago.

@oriAdler all pages extend app_layout which is exactly what makes this possible. (Implement messages once - use in all pages **if and when needed**)

Currently homepage, signin, signout, signup and User profile which will be added use those messages, without duplicating the code like you are suggesting.

This pull request has been open for 6 days, and I see no reason to delay it even more when we have 10 days left. As I said, other features rely on this feature.

oriAdler: @LeonRabinovich To be more specific, the PR for this feature was opened 8 days ago #96.
This conversation is only from yesterday, so this is not the reason for the delay.
The homepage doesn't display these messages, and we don't have a sign-out page on the application.
Anyway, as we are a team, we should make decisions together.
At the moment, you are the only one in favor of it.

LeonRabinovich: @oriAdler
Again, the homepage displays those messages after redirection..
for example, after clicking on the sign-out **button** from any page, you are redirected to the homepage, and a message is displayed for a sucsesfull sign out.(on the home page)
There is also a picture of this in one of my comments above.

Anyways, even if I disagree about the implementation, I agree that this should be a team decision.

Hence - I ask @VeraTs  and @ellaml to read this thread discussion and let me know if you think I need to change the Implemntation.

I will do as the team decides.

VeraTs: @LeonRabinovich I understand your desire for a simpler approach to the goal of this PR, given our timeline.
I even agree with it, somewhat, given that you want to display success messages as well.
However, please do note that it would not be code duplication to display it in several divs instead of in the framing, since it would serve a different purpose in each. 
For example - as to your cases, it would not be needed if in the future, sign out would lead to a "goodbye" type page instead of the front page. In case of the successful sign up - to redirect to a future "Profile" page. In both cases there would be no need to display success messages, since the pages themselves would be sufficient, and then implementation of the message display in app_layout would be redundant.
Therefore, for now, I am willing to be satisfied with the way it is now, in the app_layout even if it isn't the best approach overall.

That said, in regards to the html tags (style, head) please see what I said above, or consider deleting them and using the bootstrap for styling, since using the bootstrap classes does not require further styling and would ignore the issue altogether.

LeonRabinovich: @VeraTs 
I see your point, thank you for being satisfied with this approach for now.
Regarding the HTML tags - i will fix it and comment on the thread above when its commited.


@ellaml - i would appreciate your review on this PR as well, would help me get to the profile page PR sooner.

--
I've also added a video in the original PR comment that showcases the entire sign-in, sign-up, sign-out process, and also shows the messages that are displayed to the user.

VeraTs (approved):

homepage/templates/homepage/app_layout.html
LeonRabinovich: @VeraTs 
Fixed, thank you for the information.

homepage/templates/homepage/app_layout.html
oriAdler: @LeonRabinovich Approved, because of time constraints.
Please make sure this PR is merged ASAP.

oriAdler (approved):

homepage/templates/homepage/app_layout.html
ellaml: @LeonRabinovich Hey, sorry for taking long with my review.
I can see the upside in having this notification bar on the layout for general cases, but in my opinion code duplication is not a real problem here since each page will have its own messages, and I do prefer displaying msgs next to relevant elements for the user's actions.
Also, with that approach you maybe avoid writing similar code in two files but eventually the browser loads (due to extends) this code each time a page is loaded - any page - even if no messages will be used in it.
I would have done it differently but since we're short on time I don't think it's that important, and this notification bar can be used for other cases if needed.

ellaml (approved):

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        user = authenticate(username=not_signed_up_details[0], password=not_signed_up_details[1])
+        assert user is None
+
+    def test_signin_valid(self, signed_up_details):
+        user = authenticate(username=signed_up_details[0], password=signed_up_details[1])
+        assert user
+
+    # ------------------------Front-End testing------------------------ #
+
+    def test_post_valid_signIn_with_client(self, client, signed_up_details):
```
kobihk: python convention is func name without camel case
```suggestion
    def test_post_valid_sign_in_with_client(self, client, signed_up_details):

```

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        response = client.post('/users/sign_in/', data={
+            'username': signed_up_details[0], 'password': signed_up_details[1]})
+        HOMEPAGE_URL = '/'
+        assert response.url == HOMEPAGE_URL
+        assert response.status_code == 302
+
+    def test_invalid_signIn_redirect(self, client, not_signed_up_details):
```
kobihk: same ^^
```suggestion
    def test_invalid_sign_in_redirect(self, client, not_signed_up_details):

```

homepage/tests/sign_in_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        response = client.post('/users/sign_in/', data={
+            'username': not_signed_up_details[0], 'password': not_signed_up_details[1]})
+        assert response.url == '/users/sign_in/'
+
+    def test_renders_add_signIn_template(self, client):
```
kobihk: same ^^
```suggestion
    def test_renders_add_sign_in_template(self, client):

```

homepage/views.py
```diff
@@ -77,3 +80,34 @@ def add_review_search(request):
...
+    if request.method == "POST":
+        form = AuthenticationForm(request, data=request.POST)
+
+        if form.is_valid():
+            username = form.cleaned_data.get('username')
+            password = form.cleaned_data.get('password')
+            user = authenticate(username=username, password=password)
+
+            if user is not None:
```
kobihk: maybe you can do it like:
```suggestion
            if user:
```

kobihk (changes requested):

homepage/tests/sign_in_tests.py
LeonRabinovich: changed

homepage/tests/sign_in_tests.py
LeonRabinovich: changed

homepage/tests/sign_in_tests.py
LeonRabinovich: changed

homepage/views.py
LeonRabinovich: changed

kobihk (approved):

PR #101, opened 2021-05-09
Added single course page

### LINKED ISSUE #95: Add basic Single Course page
Add a page for a specific course with its details.
The page will have buttons for "Add Review" and "Follow",  
and will display the reviews that were posted for this course

ellaml:
- Added course page to `urls.py` that uses `course_id `as parameter in the url, and
  loads `course.html` template. When the `course_id `is not found it
  redirects to Courses page.
- Added redirection from Courses page to a specific course when clicking on its row in the table.
- Separated `single-review-box` element to a new `single_review.html`
  template in order to include it easily in other files
- Updated Reviews page to Single Course page to include
  `single_review.html` template with review as passed variable
- Added tests for valid course in url using our courses test data, and for redirection in case of invalid course

The logic for the buttons "Syllabi", "Add Review", "Follow Course" will be added later on. (They depend on other PRs)

fix: #95

### DIFF
--- a/homepage/templates/homepage/courses/course.html
+++ b/homepage/templates/homepage/courses/course.html
@@ -24,6 +24,19 @@
             </div>
             <div class="details-row">
                 <label class="course-id">{{course.course_id}}</label>
+                <div class="credit-container">
+                    <label>Credit Points:</label>
+                    <div class="credit-points">
+                        <label>{{course.credit_points}}</label>
+                    </div>
+                </div>
+            </div>
+            <div class="details-row">
+                <div>
+                    <img src="{% static 'icons/star.png' %}">
+                    <label class="avg-rating">{{course.avg_rating|floatformat}}</label>
+                    <label class="raters">({{course.num_of_raters}} Raters)</label>
+                </div>
                 <div class="course-load-container">
                     <label>Course load:</label>
                     <div class="course-load">
@@ -39,19 +52,6 @@
                         <img class="load-emoji" src="{% static 'icons/smiling.png' %}">
                         {% endif %}
                         <label>{{course.avg_load|floatformat}}/5</label>
-                    </div>
-                </div>
-            </div>
-            <div class="details-row">
-                <div>
-                    <img src="{% static 'icons/star.png' %}">
-                    <label class="avg-rating">{{course.avg_rating|floatformat}}</label>
-                    <label class="raters">({{course.num_of_raters}} Raters)</label>
-                </div>
-                <div class="credit-container">
-                    <label>Credit Points:</label>
-                    <div class="credit-points">
-                        <label>{{course.credit_points}}</label>
                     </div>
                 </div>
             </div>
--- a/homepage/templates/homepage/courses/course.html
+++ b/homepage/templates/homepage/courses/course.html
@@ -1,0 +1,75 @@
+{% extends 'homepage/app_layout.html' %}
+{% load static %}
+{% block content %}
+<html>
+<head>
+    <title>Class Rater Courses</title>
+    <link rel="stylesheet" href="{% static 'css/course.css' %}">
+</head>
+<body>
+<div class="page-container">
+    <div class="details-container">
+        <div class="details">
+            <div class="course-title-row">
+                <div class="course-title">
+                    {{course.name}}
+                </div>
+                <div class="is-mandatory-label">
+                    {% if course.mandatory %}
+                    <label>mandatory</label>
+                    {% else %}
+                    <label>non-mandatory</label>
+                    {% endif %}
+                </div>
+            </div>
+            <div class="details-row">
+                <label class="course-id">{{course.course_id}}</label>
+                <div class="course-load-container">
+                    <label>Course load:</label>
+                    <div class="course-load">
+                        {% if course.avg_load > 4 %}
+                        <img class="load-emoji" src="{% static 'icons/crying.png' %}">
+                        {% elif course.avg_load > 3.5 %}
+                        <img class="load-emoji" src="{% static 'icons/sad.png' %}">
+                        {% elif course.avg_load > 2.5 %}
+                        <img class="load-emoji" src="{% static 'icons/neutral.png' %}">
+                        {% elif course.avg_load >= 2 %}
+                        <img class="load-emoji" src="{% static 'icons/smile.png' %}">
+                        {% else %}
+                        <img class="load-emoji" src="{% static 'icons/smiling.png' %}">
+                        {% endif %}
+                        <label>{{course.avg_load|floatformat}}/5</label>
+                    </div>
+                </div>
+            </div>
+            <div class="details-row">
+                <div>
+                    <img src="{% static 'icons/star.png' %}">
+                    <label class="avg-rating">{{course.avg_rating|floatformat}}</label>
+                    <label class="raters">({{course.num_of_raters}} Raters)</label>
+                </div>
+                <div class="credit-container">
+                    <label>Credit Points:</label>
+                    <div class="credit-points">
+                        <label>{{course.credit_points}}</label>
+                    </div>
+                </div>
+            </div>
+            <div class="buttons-row">
+                <button class="course-btn">Syllabi</button>
+                <button class="course-btn">Add Review</button>
+                <button class="course-btn">Follow Course</button>
+            </div>
+        </div>
+    </div>
+    <div class="bottom-container">
+        <div class="num-review-title">{{reviews.count}} reviews available</div>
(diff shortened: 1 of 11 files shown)


ellaml:
https://user-images.githubusercontent.com/73358484/117580310-0c55d480-b100-11eb-97bb-3791f9be8a76.mp4

homepage/templates/homepage/courses/course.html
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div class="buttons-row">
+                <button class="course-btn">Syllabi</button>
+                <button class="course-btn">Add Review</button>
+                <button class="course-btn">Follow Course</button>
+            </div>
+        </div>
+    </div>
+    <div class="bottom-container">
+        <div class="num-review-title">{{reviews.count}} reviews available</div>
```
oriAdler, line 66:
```suggestion
        {% if reviews|length > 0 %}
        <div class="num-review-title">{{reviews.count}} reviews available</div>
        ...
        {% else %}
        <div class="num-review-title">no reviews available</div>
        {% endif %}
```
(optional)

homepage/tests/course_tests.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from pytest_django.asserts import assertTemplateUsed
+
+
+@pytest.mark.parametrize("valid_course_id", [
+    (10231), (10111), (10221), (12357)
+    ])
+@pytest.mark.django_db
+def test_renders_course_template(client, valid_course_id):
+    response = client.get('/course/' + str(valid_course_id) + '/')
```
oriAdler:
```suggestion
    response = client.get(f'/course/{valid_course_id}/')
```

homepage/tests/course_tests.py
```diff
@@ -0,0 +1,24 @@
...
+
+    assert response.status_code == 200
+    assertTemplateUsed(response, 'homepage/courses/course.html')
+
+
+@pytest.mark.django_db
+def test_invalid_course_resquest(client):
+    invalid_id = 11111
+
+    response = client.get('/course/' + str(invalid_id) + '/')
```
oriAdler:
```suggestion
    response = client.get(f'/course/{invalid_id}/')

```

homepage/views.py
```diff
@@ -53,3 +54,12 @@ def add_review(request):
     else:
         form = ReviewForm()
     return render(request, 'homepage/add_review.html', {'form': form})
+
+
+def course(request, id):
+    try:
+        course = Course.objects.get(pk=id)
+        reviews = Review.objects.filter(course=id)
```
oriAdler:
```suggestion
        reviews = Review.objects.filter(course=id).order_by('-avg_rating')
```
WDYT?

oriAdler (commented):
Looks Awesome as always:)
I've left a few comments

homepage/views.py
VeraTs: imo reviews should be deisplayed according to date

VeraTs (commented):
Looks great, I agree with Ori's comments, basically

homepage/templates/homepage/courses/course.html
```diff
@@ -0,0 +1,75 @@
...
+                </div>
+            </div>
+            <div class="details-row">
+                <div>
+                    <img src="{% static 'icons/star.png' %}">
+                    <label class="avg-rating">{{course.avg_rating|floatformat}}</label>
+                    <label class="raters">({{course.num_of_raters}} Raters)</label>
+                </div>
+                <div class="credit-container">
+                    <label>Credit Points:</label>
```
VeraTs: suggestion: credit points to switch places with course load in the display, so that course load and rating are on the same level?

homepage/views.py
LeonRabinovich: IMO reviews should be displayed by the number of likes as a default 
(maybe add a button to change the display by date / ratings)
Meaning the review that has the most likes will be displayed first.

I think its more helpful than displaying by avg-rating or by date, since as a user i will usually want to look at the "best" reviews first and not the newest

homepage/tests/course_tests.py
```diff
@@ -0,0 +1,24 @@
...
+from pytest_django.asserts import assertTemplateUsed
+
+
+@pytest.mark.parametrize("valid_course_id", [
+    (10231), (10111), (10221), (12357)
+    ])
+@pytest.mark.django_db
+def test_renders_course_template(client, valid_course_id):
+    response = client.get('/course/' + str(valid_course_id) + '/')
+
```
LeonRabinovich: Kobi suggested in previous reviews that we remove the blank lines before the 'assert', so i think it needs to be removed here as well

LeonRabinovich:
Looks great
I agree with Ori's review as well and added some comments

homepage/tests/course_tests.py
ellaml: @oriAdler  I tried to do it this way but it fails the test - the variables are not strings

homepage/tests/course_tests.py
ellaml: @oriAdler I tried to do it this way but it fails the test - the variables are not strings

homepage/templates/homepage/courses/course.html
ellaml: @oriAdler I'll leave it that way for now - since we need to update Reviews page as well.
We can do it as part of our "beautifying" PR next week

homepage/views.py
ellaml: Decided to sort by likes as Leon suggested

homepage/templates/homepage/courses/course.html
ellaml: switched

homepage/tests/course_tests.py
ellaml: Missed that one, if that's important in Kobi's opinion I'll update of course.
I like spaces for readability anyway :)

homepage/tests/course_tests.py
oriAdler: @ellaml my bad, the last ```/``` is redundant.
It supposed to work now. There are similar tests on PR #106

oriAdler: @ellaml There are similar tests on PR #106, also with f-Strings and client maybe it can help.
Anyway, if it doesn't work, you can leave it that way.

LeonRabinovich:
LGTM

LeonRabinovich (approved):

homepage/tests/course_tests.py
ellaml: @oriAdler Fixed. 
My bad, I had another problem.

homepage/tests/course_tests.py
ellaml: @oriAdler  same ^^

homepage/tests/course_tests.py
ellaml: I had to make another update so I fixed it on the way :)

oriAdler (approved):
LGTM

homepage/templates/homepage/reviews/single_review.html
```diff
@@ -0,0 +1,64 @@
...
+        {% for i in "123"|make_list %}
+        <img src="{% static 'icons/star.png' %}">
+        {% endfor %}
+        {% elif review.rate == 2 %}
+        {% for i in "12"|make_list %}
+        <img src="{% static 'icons/star.png' %}">
+        {% endfor %}
+        {% else %}
+        <img src="{% static 'icons/star.png' %}">
+        {% endif %}
```
kobihk: maybe you can do it simpler something like:
https://stackoverflow.com/a/17691875
```suggestion
        {% for i in range(review.rate) %}
        <img src="{% static 'icons/star.png' %}">
        {% endfor %}
```

kobihk (changes requested):

homepage/templates/homepage/reviews/single_review.html
ellaml: @kobihk I tried it but I get a template syntax error.
According to this discussion https://stackoverflow.com/questions/1107737/numeric-for-loop-in-django-templates
and this bug report https://code.djangoproject.com/ticket/5172, it looks like Django does not support it. The link you've attached is referring to Jinja.

kobihk: @ellaml please try with this code:
```suggestion
        {% for i in " "|rjust:review.rate|make_list %}
        <img src="{% static 'icons/star.png' %}">
        {% endfor %}

```

kobihk (approved):

